PR #25, opened 2016-07-27
General tidy up

BarbourSmith:


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -23,6 +23,9 @@
     */
 
 #include "MyTypes.h"
```
BarbourSmith: can this be removed and the file deleted?

PR #45, opened 2016-08-20
Pid tune with square

BarbourSmith:


cnc_ctrl_v1/axis.cpp
```diff
@@ -92,12 +92,17 @@ int    Axis::set(float newAxisPosition){
 
 void   Axis::computePID(){
     
+    if (_change(_sign(_oldSetpoint - _pidSetpoint))){
```
BarbourSmith: throw a comment on this to explain what it does

PR #62, opened 2016-11-17
Library cleanup

BarbourSmith:
move all library files to an organized structure


BarbourSmith:
test on machine, then 🚢

PR #67, opened 2016-11-29
Re write motor interface

BarbourSmith:
The GearMotor library used to use an interface which imitated the arduino Servo libarary where a commmand of 90 was stopped and 0 and 180 were full speed in opposit directions. 

This interface is clunky and doesn't provide suficient control over speed for our application so it had to be done away with. 

Rather than imitate the arduino servo library, I will write my own servo library with my interface.

### DIFF
--- a/cnc_ctrl_v1/libraries/Axis/axis.cpp
+++ b/cnc_ctrl_v1/libraries/Axis/axis.cpp
@@ -251,7 +251,7 @@ float   Axis::computeSymmetryOfMotor(int speed){
     
     //move neg for 1 sec at speed then measure dist moved
     originalEncoderPos = _encoder.read();
-    _motor.write(speed);
+    _motor.write(-1*speed);
     delay(1000);
     Serial.print('.');
     delay(1000);
@@ -310,7 +310,7 @@ void   Axis::computeBoost(){
     i                   = 0;
     originalEncoderPos  = _encoder.read();
     while (i < 35){
-        _motor.write(i);
+        _motor.write(-1*i);
         
         delay(1000);
         Serial.print(".");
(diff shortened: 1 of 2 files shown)


cnc_ctrl_v1/libraries/Axis/axis.cpp
```diff
@@ -235,30 +235,30 @@ float   Axis::computeSymmetryOfMotor(int speed){
...
+    _motor.write(0);
     Serial.print(".");
     delay(200);
     Serial.print(".");
     
     
     //move neg for 1 sec at speed then measure dist moved
     originalEncoderPos = _encoder.read();
-    _motor.write(90 - speed);
+    _motor.write(speed);
```
BarbourSmith: should be -1*speed

cnc_ctrl_v1/libraries/Axis/axis.cpp
```diff
@@ -310,7 +310,7 @@ void   Axis::computeBoost(){
     i                   = 0;
     originalEncoderPos  = _encoder.read();
     while (i < 35){
-        _motor.write(90-i);
+        _motor.write(i);
```
BarbourSmith: should be -1*i

cnc_ctrl_v1/libraries/Axis/axis.cpp
```diff
@@ -370,14 +370,14 @@ void   Axis::measureMotorSpeed(int speed){
...
-    _motor.write(90);
+    _motor.write(0);
     delay(200);
     
     //run the motor for numberOfStepsToTest steps negative and record the time taken
     originalEncoderPos  = _encoder.read();
     startTime = millis();
     while (abs(originalEncoderPos - _encoder.read()) < numberOfStepsToTest){
-        _motor.write(90 - speed);
+        _motor.write(speed);
```
BarbourSmith: should be -1*speed

PR #73, opened 2016-12-09
Fix undefined reference with new ide

BarbourSmith:
Move all the files into one folder. Not happy abou this.


BarbourSmith:
Delete folder structure

PR #84, opened 2017-01-04
Add new kinematics options

BarbourSmith:
Add kinematic options for original triangluar kinematics and new kinematics with tilt compensation


cnc_ctrl_v1/cnc_ctrl_v1.ino
```diff
@@ -31,6 +31,8 @@ void setup(){
     
     Timer1.initialize(10000);
     Timer1.attachInterrupt(runsOnATimer);
+    
+    kinematics.speedTest(.2);
```
BarbourSmith: delete

PR #94, opened 2017-01-12
Update guestBook.txt

baxerus:
Another entry in the guestbook. I was sure there would be dozens already, but nearly nothing :(

### DIFF
--- a/guestBook.txt
+++ b/guestBook.txt
@@ -12,3 +12,5 @@ Bar was back for the weekly Wednesday update!
 Kim was here too - 1/11/2017
 
 Thanks for all the great work! Thomas Kristensen (Denmark) - 12/1/2017
+
+Can't wait to get my Maslow. Keep going! Benedict (Germany) - 12/1/2017


BarbourSmith:
It's slow, but that's ok! Something I learned from my last kickstarter is that not a lot of people read the updates, but it's important to do them anyway. The video is only up to 300 views so it can't be that more than 300 people read it yet. I know at least 15 of those are me ;-)

baxerus:
How depressing :(
I kick a lot on kickstarter and sometimes I feel like I have to much project updates at the same time. Perhaps I do not read all of them instantly, because of that, but I read them. I was interested enough to back the project so of course I read them :)
Actually I am not a social media guy at all (no twitter, no facebook), but perhaps you can reach the people with updates better on this way?

By the way, I am also no explanation video guy (text is so much faster scanned for the one useful information you actually want), but I would like to see more Maslow in action. And also in details. And perhaps this time also an explanation video, how it is constructed and WHY in this way and not another. The counterweight things on the sides for example. If there are discussions about this in the forum already, linking these in an kickstarter update is also okay. Actually a small section about running discussions/topics in the community at the end of an update would be great. To encourage new members to join the discussion? Like me?

BarbourSmith:
It's OK, everyone is busy and I think that even if some people don't have time to fully read the update and go through the process of doing this they appreciate that we communicate. 

I think you are right that we should post links to the updates in more places. For the first time this week I posted a link in the forums and on Facebook. 

It's good to hear what you want to see more of. More Maslow in action is a great idea. Next week's update will be a walk-through of how to build something so we'll be cutting something out. I was trying to think of how to get more video of the machine in action without just posting boring videos of cutting and I think I will try to start doing a project every week. That way I can always be showing what I am working on in terms of software fixes, and at the same time show the machine working.

I also really like the idea of running discussion topics at the end of the update. I'll add that starting next week.

Thanks for a really solid batch of suggestions.

PR #98, opened 2017-01-14
guestbook and introduction grammar update

jbnimble:


BarbourSmith:
So cool! You found a bug in the guestbook :-) 👍 👍 👍

PR #102, opened 2017-01-17
always prompt, even for tool T1

BarbourSmith:
This is possibly not the best behavior, maybe T1 should be ignored? But
what if you want to switch back to T1 later in the program. You might
end up skipping a valid tool change


davidelang:
Have an option to skip the first tool change in the program.

David Lang

PR #120, opened 2017-02-17
Remove rapid move fix z axis behavior

BarbourSmith:


BarbourSmith:
Test that new move is as fast as old move, then merge

PR #125, opened 2017-02-21
Update chain length calibration

BarbourSmith:
Add the ability for the machine to measure the correct length of chain when it is first set up


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -421,35 +424,18 @@ void calibrateChainLengths(){
     in length
     */
     
+    Serial.println("calibrate chain lengths ran");
```
BarbourSmith: delete

PR #131, opened 2017-03-13
Fix changing machine dimensions

BarbourSmith:
Changing the machine or sled dimensions in the Ground Control settings updates the firmware to compute kinematics based on new values

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -13,7 +13,7 @@
     You should have received a copy of the GNU General Public License
     along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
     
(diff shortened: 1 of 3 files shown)


cnc_ctrl_v1/Kinematics.cpp
```diff
@@ -23,90 +23,61 @@ in X-Y space.
...
+    else if (*yTarget >  machineHeight/2){
+        *yTarget =  machineHeight/2;
     }
-    else if (*yTarget <  -MACHINEHEIGHT/2){
-        *yTarget =  -MACHINEHEIGHT/2;
+    else if (*yTarget <  -machineHeight/2){
+        *yTarget =  -machineHeight/2;
     }
     
+    //Serial.println(*xTarget);
```
BarbourSmith: 🗑

PR #133, opened 2017-03-15
Add documentation pictures

BarbourSmith:


jbarchuk:
There are often variations in how com ports behave. Folks that are used to average USB or 'friendly' devices won't have any experience with this. Here's an example. The docs will need adjustment when different problems arise. 
select-com-port-01

PR #139, opened 2017-03-22
Print ready after all commands

BarbourSmith:


BarbourSmith:
Print "ready" after each command finishes executing to make it more clear when the machine is ready

PR #143, opened 2017-03-25
Added ability to change encoder steps, dist per revolution, and gear …

johnbenweber:
…radius from serial command.

I haven't had a chance to actually test this beside running it through the Arduino compiler without error. I'm not that familiar with the inner workings of Firmware so have a look and let me know if you see any potential issues. Corresponding GroundControl modifications to come.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -26,11 +26,7 @@
 #define SIZEOFFLOAT      4
 #define SIZEOFLINSEG    17
 
-#define NUMBER_OF_ENCODER_STEPS 8148.0 
-
-
-
-Axis::Axis(int pwmPin, int directionPin1, int directionPin2, int encoderPin1, int encoderPin2, String axisName, int eepromAdr, float mmPerRotation)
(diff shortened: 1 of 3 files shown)


BarbourSmith:
Looks good. I don't see it causing issues with the rest of the firmware, but I'll test it anyway. We're going to run out of letters to send settings with soon 😄 

My only request for the Ground Control side is to maybe split the settings out into an "Advanced" section so we don't confuse everyone because most people won't want to mess with these settings.

johnbenweber:
Yeah, we can always go to double letters if that becomes an issue.

BarbourSmith:
Haha, yeah I'm not worried about it really. We can also split multiple commands if we need to.

BarbourSmith:
I haven't forgotten about this one. I'm going to merge it soon. We've had some issues with circles coming out oval so I want to make sure I merge and test the other PR which messes with settings and things which touch the kinematics thoroughly so we can keep track of what is causing what.

BarbourSmith:
I still haven't forgotten about this. I will test it and merge it today.

PR #147, opened 2017-03-26
Fix for breaking everything

BarbourSmith:
I added a watchdog that broke more or less everything. Needs to be removed right now, then fixed


BarbourSmith:
Sorry!

PR #171, opened 2017-04-13
Create placeholder gear motor encoder object

BarbourSmith:
Move the encoder object out of the axis object


cnc_ctrl_v1/MotorGearboxEncoder.cpp
```diff
@@ -0,0 +1,46 @@
...
+    Serial.println("created gear motor encoder");
+    
+}
+
+/*
+
+The encoder needs the flowing functions as pass through:
+
+-creating the encoder with pins
+-read
```
BarbourSmith: 🥔

PR #175, opened 2017-04-13
Compute motor speed

BarbourSmith:
Add a function which can compute the motors speed in real time


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -91,7 +91,7 @@ void  returnPoz(float x, float y, float z){
     int                  timeout = 200;
     
     if (millis() - lastRan > timeout){
-        
+        /*
```
BarbourSmith: 🐙

PR #185, opened 2017-04-24
fix-feed-rate-limit-in-metric-mode

blurfl:
Previous version worked incorrectly when in metric mode


BarbourSmith:
This one looks like it might have something wrong with it 😀 I'm seeing 24 files changed

BarbourSmith:
Is it ok if I close this one? It's got all the file changes and it might be faster to just make a new PR with just the lines we want 👍

blurfl:
Cancel and close this, it was handled in a separate PR

BarbourSmith:
Does that mean it's done? I think there is still an issue out for it.

blurfl:
I’ve gotta start reading these things on a larger screen :(

  This is not fixed.

BarbourSmith:
😀

blurfl:
I'll open a new PR. The fix I proposed was to dump the conversion and use the metric value. I tested in metric and inch mode, both work. Not elegant or pretty, but it corrects the problem behavior.

PR #212, opened 2017-05-11
Tune positional PID loop

BarbourSmith:
Positional PID loop is tuned to as precise as it can be with the current step size. May need to be re-tuned after adjusting step size.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -123,7 +123,9 @@ void   Axis::computePID(){
     if(_axisName[0] == 'R'){
         Serial.print(_pidSetpoint*10.0);
         Serial.print(" ");
-        Serial.println(_pidInput*10.0);
+        Serial.print(_pidInput*10.0);
+        Serial.print(" ");
+        Serial.println((_pidSetpoint*10.0) + _pidOutput/20.0);
     }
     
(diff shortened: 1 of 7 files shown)


cnc_ctrl_v1/Axis.cpp
```diff
@@ -95,12 +95,12 @@ void   Axis::computePID(){
...
     
-    if (_change(_sign(_oldSetpoint - _pidSetpoint))){ //this determines if the axis has changed direction of movement and flushes the accumulator in the PID if it has
+    if (_detectDirectionChange(_pidSetpoint)){ //this determines if the axis has changed direction of movement and flushes the accumulator in the PID if it has
         _pidController.FlipIntegrator();
     }
-    _oldSetpoint = _pidSetpoint;
     
-    //antiWindup code
+    
+    /*//antiWindup code
```
BarbourSmith: 💥

cnc_ctrl_v1/cnc_ctrl_v1.ino
```diff
@@ -17,7 +17,7 @@
 #include "TimerOne.h"
 
 void setup(){
-    Serial.begin(19200);
+    Serial.begin(115200);
```
BarbourSmith: 📖

PR #222, opened 2017-05-23
Add G38.2 gcode for auto-zero z axis

blurfl:
grbl implements G38.2 for three axes; this only implements the z axis.
G38.2 is supposed to finish at the touch point without altering the
home position, the user would use G92 for that. We don’t have G92
(beyond the scope of this :) ) so this zeroes the z axis at the touch
point. If the touch point is not reached, the ‘STOP’ command is issued
to halt further execution without altering the zero. An example of
operation would be to execute this command file:

G20
G90
G38.2 Z-.15 F1
G20
G90
M02

The probe is assumed to be connected to AUX4, which is set at the top
of CNC_Functions.h . Ground is attached to the router (bit) and AUX4 to
the touch plate. AUX4 is configured as INPUT_PULLUP and is pulled LOW
when the probe touches the plate.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -10,12 +10,12 @@
     You should have received a copy of the GNU General Public License
     along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
     Copyright 2014-2017 Bar Smith*/
-    
-    
+
+
 /*Right now this file is a catch all for functions which will be broken out into
-libraries*/
-    
-    
+  libraries*/
+
+
 #include "Axis.h"
 #include "Kinematics.h"
 #include "RingBuffer.h"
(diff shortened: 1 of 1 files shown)


BarbourSmith:
@blurfl This looks so awesome! What a cool feature! 👍 👍 👍 

From looking at the files changed it seems like all the lines are changed (even though they clearly aren't), do you have any idea what causes that?

Other than that, looks great!

davidelang:
line endings or converting tabs to spaces are common causes of this sort of 
thing.

blurfl:
I used command-T in Arduino to reformat the file 😕. Sorry, I can re-submit from an unadulterated copy?

blurfl:
'On github, you simply append the w=1 parameter to the URL for it to ignore whitespace.

Use this URL to ignore the whitespace:
https://github.com/MaslowCNC/Firmware/pull/222/files?w=1

blurfl:
I never touch github without stubbing my toe :(, but then I learn something new :)

davidelang:
it's not a good idea to make massive whitespace changes in code (without a good 
reason). It breaks tools like git blame tracing who wrote what.

It's up to bar, but I would say that it's best to redo the modifications without 
the reformatting.

If we want to reformat all the code, that should be it's own commit that makes 
no other changes.

BarbourSmith:
That w=1 trick is neat! I didn't know we could do that.

Git has a steeper learning curve than I would like but it's so powerful.

Other than the white space thing, this looks fantastic. Thanks for defining all of the Aux pins, not just the ones you are using.

Would you be comfortable re-submitting the PR without the white space changes to keep everything clean?

blurfl:
I'm firing up the editor right now, will re-submit a clean (messier :) ) version. 
 Got bit by AUX5 and AUX6 clobbering the Rx/Tx, so I only put those four in. I see that is fixed in the nex rev of the board, though 👍

blurfl:
Closing this, a de-formatted version has been submitted.

PR #223, opened 2017-05-23
G38.2 zero z axis, de-formatted

blurfl:
grbl implements G38.2 for three axes; this only implements the z axis.
G38.2 is supposed to finish at the touch point without altering the
home position, the user would use G92 for that. We don’t have G92
(beyond the scope of this :) ) so this zeroes the z axis at the touch
point. If the touch point is not reached, the ‘STOP’ command is issued
to halt further execution without altering the zero. An example of
operation would be to execute this command file:

G20
G90
G38.2 Z-.15 F1
G20
G90
M02

The probe is assumed to be connected to AUX4, which is set at the top
of CNC_Functions.h . Ground is attached to the router (bit) and AUX4 to
the touch plate. AUX4 is configured as INPUT_PULLUP and is pulled LOW
when the probe touches the plate.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -60,6 +60,12 @@ bool zAxisAttached = false;
 #define ENCODERSTEPS   8148.0
 #define ZENCODERSTEPS  7560.0 //7*270*4 --- 7ppr, 270:1 gear ratio, quadrature encoding
 
+#define AUX1 17
+#define AUX2 16
+#define AUX3 15
+#define AUX4 14
+#define Probe AUX4 // use this input for zeroing zAxis with G38.2 gcode
+
 Axis leftAxis (ENC, IN6, IN5, ENCODER3B, ENCODER3A, "Left-axis",   LEFT_EEPROM_ADR, DISTPERROT, ENCODERSTEPS);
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Awesome! I'm merging it right now.

What a cool addition. Thank you for creating it. 🎉

blurfl:
FWIW, here are pictures of the arrangement I came up with.

aux4 connector
ground clip on probe bit
touch plate in position
touch plate

davidelang:
yep, that looks about right.

I would test to see if you get enough conductivity through the case and the bearings to be able to sense contact with the connection being to the router base instead of having to clip to the bit (someone will forget to unclip it from the bit before turning on the router, guaranteed)

blurfl:
On the Ridgid router it does indeed bwork through the case, bearings and all. I actually clip to one of the screws of the z-axis mount. Other setups may^h^h^hwill vary :) . There is the touch plate between the sled and the work surface to remember as well. Someone will route through that as well.

davidelang:
I'm a bit less worried about trying to route through the touch plate compared to 
winding up the wire around the bit as it's turned on.

If you are worried about the touch plate, use a different wire to go to a holder 
for the touch plate and refuse to turn on unlees the touch plate is connected to 
it.

PR #224, opened 2017-05-24
Seed computation -> 3x faster

BarbourSmith:
Seeding the calculation of the chain lengths with the prevous values speeds up the computation 3x allowing finer granularity of steps.


BarbourSmith:
Actually just a 2x improvement, still good. This PR also bumps the maximum feedrate up to 35 ipm / 900 mm/m

PR #229, opened 2017-05-26
Send lines of gcode to the machiene faster

BarbourSmith:


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -143,6 +151,8 @@ void  _watchDog(){
     
     if (millis() - lastRan > timeout){
         
+        Serial.println("watch dog");
```
BarbourSmith: 🌮 💥

PR #231, opened 2017-05-26
Clamp value changes

seware74:
Not earth-shattering but updated the range-clamping to use compound ternary operations... this reduced branching in several of the areas they are used.

Also eliminated a couple of redundant calculations on some floats in the kinematics code.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -12,7 +12,7 @@
 
     You should have received a copy of the GNU General Public License
     along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
-    
+
     Copyright 2014-2017 Bar Smith*/
 
 /*
@@ -25,25 +25,18 @@ in X-Y space.
 
 
 Kinematics::Kinematics(){
-    
+
 }
 
 void Kinematics::_verifyValidTarget(float* xTarget,float* yTarget){
     //If the target point is beyond one of the edges of the board, the machine stops at the edge
-    
-    if (*xTarget < -machineWidth/2){
-        *xTarget = -machineWidth/2;
-    }
-    else if (*xTarget >  machineWidth/2){
-        *xTarget =  machineWidth/2;
-    }
-    else if (*yTarget >  machineHeight/2){
-        *yTarget =  machineHeight/2;
-    }
-    else if (*yTarget <  -machineHeight/2){
-        *yTarget =  -machineHeight/2;
-    }
-    
+
+    float halfWidth = machineWidth / 2.0f;
+    float halfHeight = machineHeight / 2.0f;
+
+    *xTarget = (*xTarget < -halfWidth) ? -halfWidth : (*xTarget > halfWidth) ? halfWidth : *xTarget;
+    *yTarget = (*yTarget < -halfHeight) ? -halfHeight : (*yTarget > halfHeight) ? halfHeight : *yTarget;
+
 }
 
 void Kinematics::recomputeGeometry(){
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Awesome! All of those changes are in places where speed is critical. Great job!

I will test and merge first thing tomorrow morning!

Thank you so much. 👍 👍 👍 👍

seware74:
That's probably a good idea. 
I don't expect the board dimensions will change mid stream. :) 
Move the calc to initialization and then forget about them.
-------- Original message --------From: Scott Smith <notifications@github.com> Date: 5/27/17  11:02 AM  (GMT-06:00) To: MaslowCNC/Firmware <Firmware@noreply.github.com> Cc: seware74 <seware74@gmail.com>, Author <author@noreply.github.com> Subject: Re: [MaslowCNC/Firmware] Clamp value changes (#231) 
@blurfl commented on this pull request.

In Kinematics::_verifyValidTarget, should halfWidth and halfHeight be declared externally, as the machine dimensions only change during the rare calibration operation? I don't know whether this matters much if it only gets called once per gcode, unless the file has lots of small movements as in a complex curve.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/MaslowCNC/Firmware","title":"MaslowCNC/Firmware","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/MaslowCNC/Firmware"}},"updates":{"snippets":[{"icon":"PERSON","message":"@blurfl commented on #231"}],"action":{"name":"View Pull Request","url":"https://github.com/MaslowCNC/Firmware/pull/231#pullrequestreview-40655800"}}}

blurfl:
There's a function Kinematics::recomputeGeometry() where perhaps they belong - called if the user re-does the machine calibration.

BarbourSmith:
I agree that we don't need to declare those floats every time, but I still think this is a big improvement over the if statements, so I'm going to merge it happily. 

I would gladly merge a follow up pull request moving the float declarations also. I think @blurfl is correct that the recomputeGeometry() function is the right place for them.

I tested it at all the machine's limits, and it works like a charm. Great work! 👍 💯 🎉

blurfl:
That's great. Out of interest, just where do you think are the best places to look for ways to improve timing?

BarbourSmith:
You mean in terms of speeding up how fast the code runs?

The really time sensitive stuff (computing the PID loops) runs off a timer which seems to be very consistent.

In terms of speeding up how fast the code runs I think this PR is looking in the right place. Shaving time off how fast the kinematics can compute the position is important because it runs every .1-.3mm as the machine is moving which is a lot of times. 

We're getting pretty close to where we want to be on how fast this part runs tho.

The next place that I haven't optimized at all which I think could be sped up a lot is looking at how fast a line of gcode gets taken from being a string to us knowing what the variables are so that we can start the move. That code (functions towards the bottom of CNC_Functions.h) hasn't been looked at for speed at all, and because it's dealing with strings I think it is pretty slow.

blurfl:
Great, that's just what I wanted, thanks. I agree that Arduino strings are a good place to look for optimization 😊

davidelang:
I think we still may have some math errors to figure out first (with the new 
double-PID change, are we still cutting slightly bowed lines? do we still need 
to lie to the machine about what the dimensions are? do we still cut different 
Y distances at different parts of the sheet?).

Once we do that, I think we need to go through the math code carefully and find 
what can be pre-computed, what is so near constant that it can be replaced by a 
constant, etc.

but correctness first, and I don't know if we are there yet.

PR #233, opened 2017-05-27
Move all detatching motors into single axis move()

BarbourSmith:


davidelang:
it bothers me that we need to detach motors when they aren't being moved. That 
indicates to me that running the other motors is causing enough noise on the 
sensor lines that we think the other motors have moved.

If that's the case, then it's also happening when we are moving all axis, and 
that would generate position errors.

If we have an encoder showing movement when the motor isn't moving (or any 
movement other than +1/-1/+1/-1 etc) then I think we have a problem that we need 
to address.

can you hack up a test firmware version to detect if there are pulses on the 
encoder lines when they are detached and we are running a different axis?

BarbourSmith:
It's not so much about noise on the signals as it is about stressing the motors. While they are position holding they are constantly switching direction (which you can hear as a hum). This is especially notable for the z-axis motor which is off 95% of the time. I noticed that mine was pretty much always warm. 

The other place this was an issue also during the calibration process when only one motor is being used to measure out the chain. The other motor would twitch and hum. 

I agree that ideally if the motor isn't being commanded to move it should be silent, but the only way to do that is to introduce a deadband into the PID controller which would result in some loss of accuracy.

davidelang:
a deadband is the correct thing to do. the motors have >8000 encoder counts/rev, 
so a single encoder step is ~0.0003", a far cry from our 0.015" target :-)

P.S. the encoder res is listed as 8148, but if you divide that by 200 (the gear 
ratio) it doesn't come out to an even number, are you sure that's correct? If 
it's a 10 line encoder (40 steps/rev) on the motor and a 200:1 gear ratio, that 
should be 8000 steps/rev for the output, not 8148.

BarbourSmith:
Great point, I've added a comment to the line to clarify
`#define ENCODERSTEPS   8148.0 //7*291*4 --- 7ppr, 291:1 gear ratio, quadrature encoding`

The gear ratio is actually 291:1

PR #244, opened 2017-06-07
Parameters optimizing and another abs issue

seware74:
1) On arduino, any parameter passed that is larger than  8 bits should be passed as a reference (or pointer). Params that are read-only in the called function should be qualified as const to prevent accidental modification of the referenced variable. Thusly, (almost) all of the parameters in the internal classes that are larger than a byte have been changed to a reference. Those that should be read-only (which is most) are now qualified as const. A few that were being changed locally (as a temporary variable) within the called function have been left alone, so that the called function has a local copy to modify (which is not "right" per se, but easy and still safe) Many variables were change across the classes. Testing OK on the board(without motors or encoder feedback) when tested with GroundControl and a test gcode file, but should be tested fully due to the large amount of changes.

2) There was a probable issue in Axis::error. A call to abs() was being made with encoder.read() within. Because arduino implements abs() as a DEFINE, the parameter to it is referenced 3 times. In this case it would cause encoder.read() to be executed 3 times. The evaluation would be done on the first two calls but the returned value would be from the last. Probably not the desired behavior, though I think this is only being used in debug serial messages, currently. Modified to store the return from encoder.read() and call abs() with the stored variable. Because this could have been returning a wrong value since it's implementation, it should be tested fully as well, though it also perform as expected on the board, when tested with GroundControl and a test gcode file.


blurfl:
This seems to run OK, but not sure how to test for the Axis::error change.

seware74:
Made a change to while loop in Kinematics::Inverse to eliminate some calls to _MyTrig()  and calculation of the Criteria variables on the final iteration of the loop, that appear to be unnecessary in that last iteration. (That is... the values that these calls modify are unused in the final iteration of Inverse and are unused by the function calling Inverse.) <Crikes... it was easier to spot and fix than it is to explain.>

BarbourSmith:
I'm sorry it took me so long to merge this @seware74. I was really hesitant to make any sweeping changes like this because we were seeing such strange behavior and every time something changed it seemed to break something else in unexpected ways. Now that we've figured out we were having memory issues I feel much better about making changes because I feel confident we're not going to have unexpected effects or if we do we'll know how to track them down quickly. This PR is full of excellent changes, thank you for making it! 👍 👍 👍 👍

seware74:
No worries.  I've been busy with summer stuff so little time in the code lately. 

-------- Original message --------From: BarbourSmith <notifications@github.com> Date: 7/17/17  12:51 PM  (GMT-06:00) To: MaslowCNC/Firmware <Firmware@noreply.github.com> Cc: seware74 <seware74@gmail.com>, Mention <mention@noreply.github.com> Subject: Re: [MaslowCNC/Firmware] Parameters optimizing and another abs issue
  (#244) 
I'm sorry it took me so long to merge this @seware74. I was really hesitant to make any sweeping changes like this because we were seeing such strange behavior and every time something changed it seemed to break something else in unexpected ways. Now that we've figured out we were having memory issues I feel much better about making changes because I feel confident we're not going to have unexpected effects or if we do we'll know how to track them down quickly. This PR is full of excellent changes, thank you for making it! 👍 👍 👍 👍

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/MaslowCNC/Firmware","title":"MaslowCNC/Firmware","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/MaslowCNC/Firmware"}},"updates":{"snippets":[{"icon":"PERSON","message":"@BarbourSmith in #244: I'm sorry it took me so long to merge this @seware74. I was really hesitant to make any sweeping changes like this because we were seeing such strange behavior and every time something changed it seemed to break something else in unexpected ways. Now that we've figured out we were having memory issues I feel much better about making changes because I feel confident we're not going to have unexpected effects or if we do we'll know how to track them down quickly. This PR is full of excellent changes, thank you for making it! 👍 👍 👍 👍 "}],"action":{"name":"View Pull Request","url":"https://github.com/MaslowCNC/Firmware/pull/244#issuecomment-315829692"}}}

PR #251, opened 2017-06-20
Read check sum

BarbourSmith:
Check to see if buffers are overflowing


BarbourSmith:
This pull request brings in too many changes in too many places at once

PR #256, opened 2017-07-01
Fix PWM on pin 11 for new PCB

BarbourSmith:
The timer1 library disables PWM on pin 11 which we use to drive the left motor on the new PCB (bummer). This PR fixes the issue by using the timer1 library PWM module instead of the arduino PWM module.


BarbourSmith:
Description says pin 11, actually pin 12 😒

PR #263, opened 2017-07-17
Move static strings into PROGMEM

blurfl:
Static strings can be moved to PROGMEM using the F() macro.


blurfl:
In practice, this seems to free up just over 700 bytes of dynamic ram when measured in the firmware running.

BarbourSmith:
Works great for me! I didn't even think that all of the debut/general printing strings would consume memory like that. I was mostly thinking about the _axisName strings. Great catch, great solution. A classic example of how open source makes things better. Thank you for the excellent contribution!

blurfl:
Glad you could test it. There might be further ‘heap polishing’ possible by reserving memory for the working strings like lineToReturn in RingBuffer.cpp, see:
https://learn.adafruit.com/memories-of-an-arduino/optimizing-sram#reserve-your-strings <https://learn.adafruit.com/memories-of-an-arduino/optimizing-sram#reserve-your-strings>

 <https://learn.adafruit.com/memories-of-an-arduino/optimizing-sram#reserve-your-strings>

BarbourSmith:
That's a great point about reserving memory for the strings. I bet we're fragmenting things a lot right now with all our string manipulations.

blurfl:
Printing shouldn’t matter, but decoding the gcode strings seems like a likely place to look. You would know best, though. What other strings get frequently altered?

BarbourSmith:
I think you are right that the strings around decoding the gcode are frequently altered the most. The readLine() function in RingBuffer in particular I think is probably giving us trouble because it grows a string one letter at a time by appending to it which is kinda the worst case scenario if I understand the article right

blurfl:
I think so to. Is there a limit to how long a line can be? We could reserve slightly more than that. If there is no limit, maybe we should set one?

BarbourSmith:
There isn't a limit for how long a line can be, but I agree that is a problem. We added code to truncate decimals to four places recently which will put an effective limit on how long a line can realistically be

davidelang:
don't assume that the sender will truncate the numbers, do it in the firmware if 
you want to depend on it.
kk

blurfl:
If we reserve for the string, we should guard that the string never grows beyond that. I think grbl has a line limit (70? 80?). I’ll try to find it.

davidelang:
start by setting it to the buffer size, with the send, ok protocol, things get 
really messy if you try to send more than a buffer's worth of data.

so with a 128 byte buffer, start off with message limits that size (or slightly 
larger if you want to be careful)

blurfl:
If we reserve memory, the ringBuffer.readline process needs to be protected from a lost newline which could read into it beyond the reserve.
grbl has a 50 character limit on lines; if we’re aiming for compatibility, we should honor that.

PR #267, opened 2017-07-25
Fix ZAxis Feedrate and Calculation of Total Distance

### LINKED ISSUE #266: Coordinated Move Ignoring Z-Axis Position
When retracting the z-axis is followed by an XY coordinated move, it seems that frequently, but not always, the coordinated move doesn't wait for the z-axis to complete its operation before continuing.  This results in a squiggly cut between the two points.

Here is some G-Code where I see this happening:
The relevant g-code is:
G1 Z-0.75 F35
X-0.6872 Y-0.0042
Z0.2 F30
(the coordinated move below often happens before the above command completes)
M9
G0 X-0.1116 Y0
G1 Z0.2 F35
Z-0.1875
Y0.1039 F15
G3 X-0.1138 Y0.1314 I-0.175

krkeegan:
Fixed the calculation of moveDist which if the starting depth was negative and the finishing depth was positive was returning the wrong value.

Also discovered that the hardcoded feedrate for zaxis moves of 45mm/min was too much for my motor to keep up with.  I believe this was the main cause for the incomplete retraction before moving.

Now, the feedrate can be set by Gcode, but its upper bound is limited by the rpm of the motor.  With the stock motor this results in a feedrate of about 40mm/min.  But this number will adjust based on the pitch of the zaxis setting.

Fixes #266

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -432,7 +432,7 @@ void  singleAxisMove(Axis* axis, const float& endPos, const float& MMPerMin){
     float startingPos          = axis->read();
     float moveDist             = endPos - startingPos; //total distance to move
     
-    float direction            = -1* moveDist/abs(moveDist); //determine the direction of the move
+    float direction            = moveDist/abs(moveDist); //determine the direction of the move
     
     float stepSizeMM           = 0.01;                    //step size in mm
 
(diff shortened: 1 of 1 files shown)


davidelang:
unless using g0, this should be limited to the feedrate defined by the gcode, 
and if the defined feedrate is higher than what any axis can support, the entire 
motion should be limited to that rate.

krkeegan:
Yes, that is exactly what this does.

Plus it defines the maximum feed rate at what I am able to observe the axis can support.

BarbourSmith:
This is fantastic work. Thanks for the contribution, and great job in the way you integrated it into the existing code!

PR #269, opened 2017-07-26
Fix Calculation of Angle in Arc Function

### LINKED ISSUE #268: The Arc Function used by G2/G3, Incorrectly Calculates the Angle
This is what resulted in the horrible cut that I had the other night, which you can see here:

http://forums.maslowcnc.com/t/z-axis-not-fully-retracting-before-moving/175/7

This is unrelated to the zaxis issue.

I believe the following gcode will cause an issue.  Rather than being a nice circle, I think step 12 is read wrong and we end up with a straight line from the 9 oclock position to the 7 o clock position.

1. G0 X9.175 Y-0.05
2. G1 Z-0.75
3. G1 Y0
4. G17 G3 X6.6621 Y6.3085 I-9.175
5. G1 Z-0.6875
6. G3 X6.3085 Y6.6621 I-6.6621 J-6.3085
7. G1 Z-0.75
8. G3 X-6.3085 I-6.3085 J-6.6621
9. G1 Z-0.6875
10. G3 X-6.6621 Y6.3085 I6.3085 J-6.6621
11. G1 Z-0.75
12. G3 Y-6.3085 I6.6621 J-6.3085
13. G1 Z-0.6875
14. G3 X-6.3085 Y-6.6621 I6.6621 J6.3085
15. G1 Z-0.75
16. G3 X6.3085 I6.3085 J6.6621
17. G1 Z-0.6875
18. G3 X6.6621 Y-6.3085 I-6.3085 J6.6621
19. G1 Z-0.75
20. G3 X9.175 Y0 I-6.6621 J6.3085
21. G1 Y0.025
22. G19 G3 Y0.05 Z-0.725 K0.025
23. G1 Z0.2
24. G17

After relearning some trig, I think I spotted the problem.  I want to confirm that the error is reproducible, which if I am right it is and that my fix solves the issue.  I should be able to do that tonight.

krkeegan:
The calculation of theta was wrong.  A good example was theta would equal zero for any cut in quadrant 2 ending in quadrant 3, where the start and end angles were the same.

To fix:
1. flip the calculation so it is ending point minus starting point.
2. The values of atan2 are between -pi and pi.  So we need to normalize to 2pi

No longer need to specifically catch the edge case of a full circle as this is now caught as part of the regular processing.

Fixes #268

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -631,11 +631,17 @@ int   arc(const float& X1, const float& Y1, const float& X2, const float& Y2, co
     float endingAngle            =  atan2(Y2 - centerY, X2 - centerX);
     
     //compute angle between lines
-    float theta                  =  abs(startingAngle) - abs(endingAngle);
-    
-    //Catch the corner case where the beginning and end of the circle are the same
-    if (startingAngle == endingAngle){
-        theta = direction*2*pi;
+    float theta                  =  endingAngle - startingAngle;
+    if (direction == COUNTERCLOCKWISE){
+        if (theta <= 0){
+            theta += 2*pi;
+        }
+    }
+    else {
+        //CLOCKWISE
+        if (theta >= 0){
+            theta -= 2*pi;
+        }
     }
     
     float arcLengthMM            =  circumference * (theta / (2*pi) );


krkeegan:
That one was a rough one for me.  It has been a long time since I did any real trig and I don't think I have ever used atan2 before.

Anyways, I was able to reproduce the error both clockwise and counterclockwise.  I have sample files if anyone else wants to try.  After this change, no more errors.

BarbourSmith:
Love it! Great work!

I waited to merge until after the weekly release just so I would have more time to test. These trig things can be tricky with all sorts of weird corner cases

blurfl:
Another good, solid contribution! 💯

PR #271, opened 2017-07-27
Reduce string memory use

BarbourSmith:
Reserve space for strings to prevent stack fragmentation


BarbourSmith:
@dlang this is the branch I mentioned in the forums

PR #272, opened 2017-07-27
Add Bxx codes for heap inspection

blurfl:
B12 executes Avrheap.freeListWalk
B13 executes Avrheap.dumpHeap
Both require AvrHeap library from:
https://github.com/RobTillaart/Arduino/tree/master/libraries/AvrHeap

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -22,6 +22,10 @@ libraries*/
 
 #define VERSIONNUMBER 0.82
 
+#include <Avrheap.h>
+
+Avrheap myheap;
+
 bool zAxisAttached = false;
 
 #define FORWARD           1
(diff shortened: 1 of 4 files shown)


blurfl:
These put heap info into the log. Unfortunately they didn't show anything that looked useful to my eyes - YMMV.

BarbourSmith:
I'm going to take a look now.

Not seeing anything might just mean we're looking in the wrong place which would be valuable 👍 👍 👍

BarbourSmith:
This is fantastic! I agree that I don't see any noticeable changes to he heap as the issue progresses, but it's so cool to be able to look at the heap directly and see how it changes. Great work!

blurfl:
How about a BXX code to make changes to the PID values?

BarbourSmith:
I like it. I also think you are spot on that the PID values are next place to look. By setting ` _Ki = 0` in MotorGearboxEncoder.h I was able to make the issue go away.

How do you feel about putting the option to change the PID values in the advanced settings instead of as a Bxx code?

blurfl:
Very interesting! Advanced settings would be a good place for them.

blurfl:
Thinking further about this, I think it would be good to include a “reset to default values” button in the same place. Explorers need an escape plan ☺️...

BarbourSmith:
I've been thinking exactly the same thing! 😀 

I think I might make a forum thread to talk the idea through and get more opinions

PR #276, opened 2017-07-28
Catch Rapid Movement When Gcode in Prior Line

krkeegan:
Rapid movements were not being executed as rapid when the G0 code was on a prior line.

Also remove unused variable from G1 function.


BarbourSmith:
Great catch! Thank you! I will test and merge this today

BarbourSmith:
Looks great! Thank you for the contribution!

PR #278, opened 2017-07-28
Delay After Each Step, Not Before; Remove Unneeded Variables

krkeegan:
The delay should occur after each step has taken place to allow the final step to be occur before the next instruction.  Similarly no delay is needed before executing the first step of an instruction.  The while loop was instead causing the delay to occur before issuing the first movement and not one after the final movement.

Since these delays generally cancel out, I doubt there is much improvement in precision.  Any such improvement would be on the order of sub millimeter anyways.

Also remove unneeded beginingOfLastStep Long variable.  This isn't needed if we use Delay.


BarbourSmith:
Great point! This does seem like a better way to do it. Thank you! 👍 👍 👍

PR #281, opened 2017-07-31
removes the .DS_Store file and updates the .gitignore for repo hygene

designfrontier:
Just removes an OS X specific cache file from the repo and adds the globs to prevent it or any others like it from showing up.

https://en.wikipedia.org/wiki/.DS_Store


BarbourSmith:
Great! Thank you!

PR #285, opened 2017-08-07
changed _pid variables to volatile

seware74:
_pidSetpoint _pidInput and _pidOutput should be volatile since they are modified and accessed by  multiple execution contexts (ISR and mainline code) to ensure that the value is retrieved from RAM and not a register

### DIFF
--- a/cnc_ctrl_v1/Axis.h
+++ b/cnc_ctrl_v1/Axis.h
@@ -61,7 +61,9 @@
             int        _currentAngle;
             int        _previousAngle;
             double     _timeLastMoved;
-            double     _pidSetpoint, _pidInput, _pidOutput;
+            volatile double _pidSetpoint;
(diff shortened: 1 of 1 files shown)


davidelang:
given that there is only a single core on the current devices, does this matter?

It's not a bad thing to make it clean for future multi-core devices, but I'm 
trying to understand if you think this could be the cause of any current bugs.
k

seware74:
yes absolutely... these variables are modified and used in multiple execution contexts. Any time data is shared between an ISR and the main loop() they should be declared as volatile. Cores are the not the only way to introduce multiple executions contexts. ISRs are the most elementary, of which we have a quite complicated one.  In lieu of more involved solutions (e.g. critical sections or disabling interrupts), this is a bit of quick insurance.

BarbourSmith:
👍 👍 👍 👍

PR #287, opened 2017-08-17
Split b03 into multiple commands

BarbourSmith:
We were seeing issues with the kinematics being unable to load the machine's position.

The issue was that if the machine tried to load it's position from memory before it had been told the specs for the motors it would load an incorrect position, then save incorrect chain lengths to memory. The next time the position was loaded from memory it would be wrong and the machine would try to compute a new position from the incorrect chain lengths and fail

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -913,7 +913,7 @@ void  calibrateChainLengths(){
     Serial.print(rightAxis.read());
     Serial.println(F("mm"));
     
-    //kinematics.forward(leftAxis.read(), rightAxis.read(), &xTarget, &yTarget);
+    kinematics.forward(leftAxis.read(), rightAxis.read(), &xTarget, &yTarget);
     
     
     leftAxis.setPIDAggressiveness(1);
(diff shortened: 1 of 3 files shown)


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -947,6 +945,62 @@ void  updateSettings(const String& readString){
...
+    
+    //propagate the new values
+    kinematics.recomputeGeometry();
+    
+    Serial.println("leftAxis.read:");
+    Serial.println(leftAxis.read());
+    
+    kinematics.forward(leftAxis.read(), rightAxis.read(), &xTarget, &yTarget);
+    
+    Serial.println(F("Kinematics Settings Updated"));
```
BarbourSmith: delete

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -947,6 +945,62 @@ void  updateSettings(const String& readString){
...
+        kinematics.kinematicsType = kinematicsType;
+    }
+    if (rotationDiskRadius != -1){
+        kinematics.rotationDiskRadius = rotationDiskRadius;
+    }
+    
+    //propagate the new values
+    kinematics.recomputeGeometry();
+    
+    Serial.println("leftAxis.read:");
```
BarbourSmith: delete

PR #292, opened 2017-08-31
Require Machine Settings Before Unsafe Commands

### LINKED ISSUE #291: EEPROM Data Could be Written Before Machine Settings are Received
If the main loop is run without receiving any machine data, it will save the axis positions to the EEPROM as 0.0, overwriting any prior valid data.

krkeegan:
Requires that both the B12 and B03 commands be received prior to executing any unsafe commands.  A whitelist of safe commands (those that can safely be executed without machine data) is included and must be added to if new safe commands are added in the future.

Also swiches to get/put/update EEPROM commands which saves us from having custom made read/write float commands as well as ensures that writes to the EEPROM are only done when necessary.

Finally, gracefully handles receiving empty commands.

This solves a major bug in which EEPROM data could be overwritten by other processes on the host computer.
Fixes #291

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -140,6 +140,12 @@ float _inchesToMMConversion =  1;
 bool  useRelativeUnits      =  false;
 bool  stopFlag              =  false;
 bool  pauseFlag             =  false;
+bool  rcvdKinematicSettings =  false;
+bool  rcvdMotorSettings     =  false;
+bool  encoderStepsChanged   =  false;
+bool  zEncoderStepsChanged  =  false;
+// Commands that can safely be executed before machineReady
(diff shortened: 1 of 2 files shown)


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -1052,6 +1096,11 @@ void  executeGcodeLine(const String& gcodeLine){
     
     */
     
+    if (!machineReady() && !isSafeCommand(gcodeLine)){
+        Serial.println(F("Unable to execute command, machine settings not yet received"));
```
BarbourSmith: Thanks for putting in an error message for when commands are ignored 👍 👍

BarbourSmith:
This looks fantastic! Great work @krkeegan!

I'm away from my machine at a good friend's wedding until Monday, is anyone else able to test this PR on hardware? If not I can still test and merge it before next week's update.

BarbourSmith:
Tested perfect on my machine! Excellent addition @krkeegan! Thank you! 👍 👍 ❤️ ❤️

PR #294, opened 2017-09-06
RingBuffer updates, add Spindle Power control, add verboseDebug, update G-code handling, update watchdog, add support for Tool Change popup

wrh3c:
Please review my commits for possible inclusion in the Maslow firmware.

Highlights:
- Add RingBuffer::numberOfLines() to count the number of full lines in the buffer, other minor updates to RingBuffer.
- Add Spindle Power control using discrete output or servo, with method and parameters hard-coded for now until they can be integrated in settings.
- Fix G-code lastCommand to remember only G00 through G03 for when future lines omit the G number.  (It doesn't make sense to remember other G numbers as they are not movement commands.)
- Change _watchdog to request new code from Ground Control when necessary, but never add an end-of-line character to the buffer - this is how the buffer gets filled with extra lines and eventually overflows.
- Update tool change handling to pause on M06 so the user can change the tool.  Requires Ground Control update to add ToolChangePopup.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -1347,6 +1347,11 @@ void  executeGcodeLine(const String& gcodeLine){
     
     */
     
+    if (!machineReady() && !isSafeCommand(gcodeLine)){
+        Serial.println(F("Unable to execute command, machine settings not yet received"));
+        return;
+    }
+    
     //Handle G-Codes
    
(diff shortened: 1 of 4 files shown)


blurfl:
There are several very good improvements here, but lumping them together means that a problem with any one will require rolling back all of them. It’s better to keep PRs to single issues or items.

wrh3c:
Sorry, I'm still learning how to use GitHub.  I understand the issue with lumping these together now.  I'll close this pull request and submit separate ones.

PR #295, opened 2017-09-06
RingBuffer updates and add numberOfLines

wrh3c:
- In RingBuffer::read(), move _incrementBeginning() inside else statement.  If the buffer is empty, we definitely don't want to move the beginning pointer.
- Add RingBuffer::numberOfLines() to count the number of full lines in the buffer, update RingBuffer::readLine() to use this count instead of looking for a full line.
- Update RingBuffer::print() to dump the entire buffer contents, update RingBuffer::_incrementEnd() to use RingBuffer::print() if there is a buffer overflow to print buffer info and contents instead of using its own separate code.
- Fix comments for RingBuffer::spaceAvailable() and RingBuffer::length().

### DIFF
--- a/cnc_ctrl_v1/RingBuffer.cpp
+++ b/cnc_ctrl_v1/RingBuffer.cpp
@@ -54,10 +54,30 @@ char RingBuffer::read(){
     else{
         letter = _buffer[_beginningOfString];     //else return first character
         _buffer[_beginningOfString] = '\0';       //set the read character to null so it cannot be read again
+        _incrementBeginning();                    //and increment the pointer
     }
-    _incrementBeginning();
     
     return letter;
+}
+
+int RingBuffer::numberOfLines() {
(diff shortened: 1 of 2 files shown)


wrh3c:

cnc_ctrl_v1/RingBuffer.cpp
```diff
@@ -54,12 +54,32 @@ char RingBuffer::read(){
...
+    /*
+   
+    Return the number of full lines (as determined by \n terminations) in the buffer
+   
+    */
+
+    int lineCount = 0;
+
+    int  i = _beginningOfString;
+    while (i !=  _endOfString) {    // if we haven't gotten to the end of the buffer yet
```
BarbourSmith: I really like your commenting style. Very clear and to the point, but nothing is left out of the comments either 👍

PR #297, opened 2017-09-06
Watchdog - never add EOL to buffer

wrh3c:
Change _watchdog() to request new code from Ground Control when necessary, but never add an end-of-line character to the buffer - this is how the buffer gets filled with extra lines and eventually overflows.

Also change timeout to an unsigned long to make sure comparison with millis() and lastRan is successful.


wrh3c:

PR #299, opened 2017-09-08
Reduce chance of ring buffer overflow - _signalReady

wrh3c:
Now that we have the updated RingBuffer with numberOfLines() from pull request #295, here is a suggested improvement to reduce the change of ring buffer overflow.  This prevents _signalReady() from sending "ok" to Ground Control if there are 2 or more lines already in the buffer.  This is probably not necessary since pull request #297 was also merged, but it shouldn't hurt, and it might be helpful if future changes inadvertently start putting additional lines in the buffer.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -226,7 +226,9 @@ void  _signalReady(){
     
     */
     
-    Serial.println(F("ok"));
+    if (ringBuffer.numberOfLines() < 2) {         // if there are fewer than 2 lines in the buffer
+        Serial.println(F("ok"));                  // request new code
+    }
 }
 
 void  _watchDog(){
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -135,6 +135,8 @@ Axis zAxis    (ENB, IN3, IN4, ENCODER2B, ENCODER2A, "Z",     Z_EEPROM_ADR);
 Kinematics kinematics;
 RingBuffer ringBuffer;
 
+int expectedMaxLineLength   = 60;   // expected maximum Gcode line length in characters, including line ending character(s)
+
 float feedrate              =  500;
 float _inchesToMMConversion =  1;
 bool  useRelativeUnits      =  false;
@@ -226,8 +228,9 @@ void  _signalReady(){
     
     */
     
-    if (ringBuffer.numberOfLines() < 2) {         // if there are fewer than 2 lines in the buffer
-        Serial.println(F("ok"));                  // request new code
+    if ( (ringBuffer.spaceAvailable() > expectedMaxLineLength)    // if there is space in the buffer to accept the expected maximum line length
+          && (ringBuffer.numberOfLines() < 4) ) {                 // and if there are fewer than 4 lines in the buffer
+        Serial.println(F("ok"));                                  // then request new code
     }
 }
 


wrh3c:
Bar, this brings up a question I have about the purpose of the ring buffer.  Is the firmware trying to buffer a line or two ahead of the one the machine is currently cutting, or is this just to buffer the characters sent by Ground Control until the firmware has a complete line that can be processed?

davidelang:
On Thu, 7 Sep 2017, William Hall wrote:

> Bar, this brings up a question I have about the purpose of the ring buffer.  Is the firmware trying to buffer a line or two ahead of the one the machine is currently cutting, or is this just to buffer the characters sent by Ground Control until the firmware has a complete line that can be processed?

yes :-)

If there is some monster long line, we need to make sure we get it all.

we also have data arriving unpredictably from the host, and need to catch it 
with the least overhead possible so that we don't miss some.

and if we wait to finish processing one line before telling the host to send the 
next line, the machine can stall and become jerky. We ideally want much more 
than two lines in our queue, if someone is sending a bunch of points that only 
differ in position a couple points past the decimal place, it takes as long to 
interpret the command than to execute it, even before factoring in the delay for 
the host to detect the Ok and decide to send the next line.

David Lang

wrh3c:
Okay, we should consider how to implement this.  In my experience, Ground Control sends one line in response to an "ok".  Since the firmware only sends "ok" after completion of a line, there is only one line in the ring buffer.  Should we have Ground Control initially send several lines before waiting for the first "ok" response?  Or should we have the firmware send multiple "ok"s until the buffer has several lines waiting?

The ring buffer size is currently only 128 characters, so there isn't a lot of room for multiple long lines.

wrh3c:
I added a second commit to this pull request to revise my original commit.  Now _signalReady() first checks that there is enough space available in the buffer before requesting another line of code, then checks that there are fewer than 4 lines in the buffer.

I made expectedMaxLineLength a global variable rather than a constant in case we want to add code in the future to adaptively determine the expected maximum line length, or let the user change it at run time through Ground Control.

BarbourSmith:
Fantastic!

Here's the back story on how this system was designed to work, and how it actually does. 

It's based on the GRBL standard of character counting. The idea is that the computer keeps track of how many characters it has sent and when the buffer is full (because the buffer size is known) it stops sending them. Here is their wiki article on how it works: https://github.com/grbl/grbl/wiki/Interfacing-with-Grbl

It worked pretty well, but sometimes (maybe 1/500) lines would drop a character from the front. We couldn't figure out what was going on so we toned things back to the current system where the next line isn't sent until the last one is executed. Now the system is stable, but the delay is noticeable in files with many many small line segments so I would ultimately like to go back to character counting (or something similar).

Personally I'm not a huge fan of the idea of character counting. I also played around with a system where the buffer reported it's available space directly to Ground Control but never got it to work quite right

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -22,6 +22,11 @@ libraries*/
 
 #define VERSIONNUMBER 0.89
 
+#define verboseDebug 2    // set to 0 for no debug messages, 1 for single-line messages, 2 to also output ring buffer contents
+
+#include <Servo.h>
+Servo myservo;  // create servo object to control a servo
```
BarbourSmith: Why is there a servo involved? 😀

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -244,7 +256,10 @@ void  _watchDog(){
         if (!leftAxis.attached() and !rightAxis.attached() and !zAxis.attached()){
             
             if (ringBuffer.length() == 0) {       // if the buffer is empty
-                _signalReady();                          //request new code
+                #if defined (verboseDebug) && verboseDebug > 0
```
BarbourSmith: 👍 for #define verbose!

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -1072,6 +1104,9 @@ void updateMotorSettings(const String& readString){
     Serial.println(F("Motor Settings Loaded"));
 }
 
+// *** There is a more elegant way to do this - put the machineReady check at the beginning of each non-safe code!
```
BarbourSmith: ❤️

cnc_ctrl_v1/CNC_Functions.h
BarbourSmith: Oh! I get it! It's just and easy way to get a PWM output! 

The servo library might be affected by some of the timers we're using for computing the PID loops, just something to keep an eye on

wrh3c:
Hmm, I didn't mean for all of these commits to get merged.  I thought I was being careful to keep things in separate branches, but something went wrong.  Let me see where things ended up.

cnc_ctrl_v1/CNC_Functions.h
wrh3c: Actually, I'm experimenting with using a servo to physically turn a switch on and off to control power to the router.  I was concerned about possible impact to timers which is why I wanted to test this first.  For minimal impact, I attach() the servo, use it, then detach() it, so hopefully there is no impact to timers.  I didn't see an issue when I ran this on my Maslow two days ago.

I also included the option to use a digital I/O pin to control a relay.  I know you're reluctant to have the Maslow control power to the router, so I wasn't planning on rolling this out just yet.  But if we make the defaults "controlPin - AUX1", "useServo = false", and "delayAfterChange = 0", there should be no time penalty, and no impact as long as AUX1 isn't connected to anything.

wrh3c:
I added a bunch of comments in pull request #300 about the extra stuff that came along with this merge.

wrh3c:
Bar, thanks for the back story on how this was designed to work, and for the GRBL reference.  Very interesting reading!

As for the dropped character you occasionally experienced with character counting, it may have been due to a difference in counting between Ground Control and the Firmware.  If Ground Control miscounts and sends one extra character, causing the ring buffer to overflow, the extra character will replace the previous character at the end of the buffer.  RingBuffer::write() puts the character at the end of the string, then increments the end position, but _incrementEnd() checks for buffer overflow before incrementing the end position.  Even if RingBuffer::write() checked for buffer overflow and didn't replace the character at the end of the buffer, the character passed to RingBuffer::write() would be lost.

wrh3c:
The GRBL look-ahead planner buffer sounds like a reasonable solution.  Instead of storing multiple lines in the ring buffer, take one line, parse it, store it in the look-ahead planner buffer, and return "ok" so the next line can be sent.  Then when the look-ahead planner buffer gets full, hold the "ok" response until there is room for the current line in the ring buffer.

I think the main reason for the ring buffer is to store normal characters (Gcode) so that the immediate characters (such as "~", "!") can be acted upon as soon as they are received.  Otherwise, the immediate characters would be stuck in the serial receive buffer and would not have priority over the normal characters.

davidelang:
one thing to keep in mind is that on the arduino there is no flow control, so if 
the buffers fill up, the sender has no way of finding out.

PR #300, opened 2017-09-08
Add maslowDelay

wrh3c:
Adds a maslowDelay() function to allow the firmware to delay for a certain amount of time while maintaining communication with Ground Control.

Also replaces the wait loop in G1() with a call to maslowDelay().  The wait loop in G1() does not continue communication with Ground Control, which can cause Ground Control to think that the connection is lost.  This can be illustrated as follows:
- Change waitTimeMs in the wait loop in G1() from 1000 (1 second) to 10000 (10 seconds) and upload to the Arduino
- Run Ground Control and turn off "z-axis installed" in settings
- Run a Gcode file in Ground Control
- After the first manual z-axis adjustment, Ground Control will complain that the connection has been lost

To illustrate that this fix works, after making the change to use maslowDelay() in G1(), change the argument passed to maslowDelay() from 1000 (1 second) to 10000 (10 seconds) and run Ground Control as above.  There is a much longer delay after each manual z-axis adjustment, but the Arduino still communicates with Ground Control and the connection is maintained.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -320,6 +320,26 @@ void pause(){
     }    
 }
 
+void maslowDelay(unsigned long waitTimeMs) {
+  /*
+   * Provides a time delay while holding the machine position, reading serial commands,
+   * and periodically sending the machine position to Ground Control.  This prevents
+   * Ground Control from thinking that the connection is lost.
+   * 
+   * This is similar to the pause() command above, but provides a time delay rather than
+   * waiting for the user (through Ground Control) to tell the machine to continue.
+   */
+   
+    unsigned long startTime  = millis();
+    
+    while ((millis() - startTime) < waitTimeMs){
+        delay(1);
+        holdPosition();
+        readSerialCommands();
+        returnPoz(xTarget, yTarget, zAxis.read());
+    }
+}
+
 bool checkForProbeTouch(const int& probePin) {
   /*
       Check to see if AUX4 has gone LOW
@@ -602,14 +622,8 @@ int   G1(const String& readString, int G0orG1){
             pause(); //Wait until the z-axis is adjusted
             
             zAxis.set(zgoto);
-            
-            int    waitTimeMs = 1000;
-            double startTime  = millis();
-            
-            while (millis() - startTime < waitTimeMs){
-                delay(1);
-                holdPosition();
-            } 
+
+            maslowDelay(1000);
         }
     }
     


krkeegan:
maslowDelay is certainly needed.  There are a number of places where the code makes very similar calls to check for a stop command, I know there is one in arc() as well.

It makes sense to unify this under a single function.

It looks like there is a bunch of other changes in here too, are these other pull requests?  Such as setSpindlePower() and such?

BarbourSmith:
It looks fantastic! I agree with @krkeegan that maslowDelay is a both needed and well executed improvement.

I'm also seeing a lot of other changes, but I think that's just because this branch also has all the changes in #299 . When #299 is merged it should tidy up automatically, right?

Any objections to me just merging both right now?

krkeegan:
#299 actually contains some of the structure of maslowDelay().  I am not sure you can merge #300 without #299 or vice versa.

This also includes a new function setSpindleSpeed().

I would like to avoid being pedantic, so if there has been enough testing to ensure that this works, then I think these should both be merged.

But it would be helpful in the future to try and make merges more concise.  Not only are they easier to read, but it makes it easier to resolve merge conflicts when merging these in to other forks.

wrh3c:
Hmm, this was only supposed to be the "Add maslowDelay" commit.  The rest are works in progress that I'm still developing.  I thought I was careful to keep them in a separate branch, but apparently I messed something up.

krkeegan:
That makes sense.

So when you make a pull request, the pull is for the branch _at that state_ not a separate bucket of only the commits uniquely made on that branch.

I often find myself in a similar situation where I realize I have a bunch of commits that should probably be split up.  What I do in that case, is checkout upstream/master and create a new branch from it.  I then use git-cherry-pick to copy over the relevant commits and make a pull request for this branch only.

Of course, you have to be careful to ensure that all of the relevant work gets bundled up properly.

wrh3c:
Yes, I was careful to start with a clean copy of the master, created a new branch, made only the desired change, and verified when I submitted the pull request that only the one commit was included.  However, it appears that GitHub will automatically update the pull request if other commits are made in the branch.  That happened to me in another pull request, but in that case, it was the desired behavior.  I'm not sure what happened here, because I was working in a different branch today, but somehow other commits got added to this one.  From now on, I'm going to do my work outside of the GitHub-controlled directory and pull in just what I want when I want.

wrh3c:
The good news is that all changes were confined to CNC_Functions.h, so as a worst case, we could revert back to the previous version.

However, if we stick with the current version, the good news for me is that it's very close to my "work in progress" version, so it will be much easier to suggest changes going forward.

The current version is almost identical to the version that I tested two days ago actually cutting wood on my Maslow.  The only difference is the added condition to check ringBuffer.spaceAvailable in the if statement in _signalReady().  So it has been tested to that extent.

davidelang:
what you need to do is to create a separate branch for each feature, branching 
off from the last release.

Then when you get ready to create the PR, create a new branch, merge in any 
updates to master, and find any fixes you need, then delete this test branch and 
create a PR from the feature branch

try not to create one feature branch from another one (unless the second feature 
depends on the first one)

as you found, a PR submits the entire branch, not just particular commits.

to go from a combined branch, you will need to create new brances from master, 
and cherry-pick commits from the combined branch (and watch out that you don't 
suck in more changes than you think)

Remember, git tracks the state of the files, not a list of changes.

wrh3c:
Bar, feel free to ask questions and push back before accepting my pull requests.  I really didn't intend for all of this to come through in one request.

But now that it did, here's what you got:
- #define verboseDebug is added to easily enable or disable debug messages at compile time.
- setSpindlePower() to turn the spindle (router) on or off using a digital output to control either a relay or a servo, where the servo physically turns a switch on or off.  The digital output state can be chosen as active high or active low, and the servo angles can be specified for idle, on, and off positions.
- executeGcodeLine() is split into three separate functions: executeBcodeLine(), executeGcodeLine(), and executeOtherCodeLine().  There is also a new executeMcodeLine() to handle spindle power, tool changes, and program pause and end commands.
- interpretCommandString() is updated to call the separate functions mentioned above, and searches for 'G' or 'M' codes to execute in order from left to right.

krkeegan:
So something isn't working for me.  I am seeing the B12 setup command properly read, but the B03 command is stuck in the ring buffer, for some reason it doesn't seem to think that it has received an EOL.

wrh3c:
Do you have #define verboseDebug 2 set in the firmware?  If so, take a look at the log.txt file after quitting Ground Control.  You should see "Sent: B12 ...", and later you should see the B12 string after "Buffer Contents:".  The B12 string should end with "\n" just before "(End of Buffer)".  Then you'll see "iCS executing B code line:" with the B12 string.  (iCS is an abbreviation for interpretCommandString.)

Further down, you should see "Sent: B03 ...", then "Buffer Contents:" followed by the B03 string ending with "\n".  Then you should see "iCS executing B code line:" with the B03 string.  I see this, but based on your comment that it's stuck, I'm guessing you won't see it.

If you attach your log.txt file, I'd be happy to take a look and see if I can figure out what's happening.

PR #308, opened 2017-09-22
Upgrade PID Library to v1.2.1

### LINKED ISSUE #304: Upgrade PID Library to 1.2.1
This release adds a new feature call Proportional on Measure PonM which you can read about:

http://brettbeauregard.com/blog/2017/06/introducing-proportional-on-measurement/
http://brettbeauregard.com/blog/2017/06/proportional-on-measurement-the-code/

So, from my basic understanding.  The benefits of PonM are most applicable to systems where the output affects the rate of change of the input.  Heating elements seem to be the canonical example.

In our system, the velocity controller seems to be largely proportional to me.  The output generally seems to directly affect the input.  And in playing with the controller, I can largely get good results with little overshoot.  Plus mild overshoot in velocity may actually be a slight benefit.

The position controller however is another thing, its output velocity affects the rate of change of the input, position.  I suspect that this controller would greatly benefit from this feature.

krkeegan:
This does not alter the functionality of the library nor how it is used in Maslow.  Specifically the proportional aspect of the PID calculation is still handled the same way it always has been.

Fixes #304

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -31,7 +31,7 @@ Axis::Axis(const int& pwmPin, const int& directionPin1, const int& directionPin2
 motorGearboxEncoder(pwmPin, directionPin1, directionPin2, encoderPin1, encoderPin2)
 {
     
-    _pidController.setup(&_pidInput, &_pidOutput, &_pidSetpoint, 0, 0, 0, REVERSE);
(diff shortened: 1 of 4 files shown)


BarbourSmith:
Looks great to me! 👍 👍 

It's cool to be able to see all the changes to the library highlighted like this.

I hadn't seen the 

> Fixes #304 

type of notation before where it auto closes the issue. Very cool. Great work.

PR #309, opened 2017-09-22
Don't Run PID Calculation if Motors are Detached/Unpowered

### LINKED ISSUE #302: Is it Appropriate to Calculate PID When Axes are Detached?
I have no experience with PID calculations, but is it normal to continue calculating the PID values when the controlled device is not active?

The P and the D terms don't look like they would be affected.  But the integral term continues to increase/decrease until it reaches min/max. I suspect that this causes a bit of a jump whenever the axes are reattached after being detached.

We don't generally detach the axes in the middle of a cut, so this likely doesn't arise much.  I can think of a few cases:

1. Likely mostly on the very first move.
2. I think if a zaxis move takes more than 2 seconds, there is the potential that the left and right axes could be detached briefly.  
3. I _think_ if there were a sequence of zaxis moves, such as a plunge and then retract, that the left and right axes could be disabled for the entire axis retract time.
4. Finally, whenever we click hold.

https://github.com/MaslowCNC/Firmware/blob/7cbfecb67fa7288fa00368b069425aba0000a289/cnc_ctrl_v1/cnc_ctrl_v1.ino#L42-L43

https://github.com/MaslowCNC/Firmware/blob/7cbfecb67fa7288fa00368b069425aba0000a289/cnc_ctrl_v1/Axis.cpp#L99-L115

https://github.com/MaslowCNC/Firmware/blob/7cbfecb67fa7288fa00368b069425aba0000a289/cnc_ctrl_v1/PID_v1.cpp#L51-67

krkeegan:
This was causing the integrator term to wind up while the machine was on, but not running.

Fixes #302

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -98,8 +98,7 @@ int    Axis::set(const float& newAxisPosition){
 
 void   Axis::computePID(){
     
-    
-    if (_disableAxisForTesting){
+    if (_disableAxisForTesting || !motorGearboxEncoder.motor.attached()){
         return;
     }
     


BarbourSmith:
Looks great to me!

PR #311, opened 2017-09-25
Allow Weighting of Proportional on Error/Measurement

krkeegan:
Enables weighting of the proportional calculation method.  1 = proportional on error, 0 = proportional on measurement.  Any fraction in between balances the two options.

Code was a suggestion from the author of the PID library:
http://brettbeauregard.com/blog/2017/06/proportional-on-measurement-the-code/

Must be merged with changes to Ground Control as well, https://github.com/MaslowCNC/GroundControl/pull/426.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -126,7 +126,7 @@ void   Axis::computePID(){
     
 }
 
-void   Axis::setPIDValues(float KpPos, float KiPos, float KdPos, float KpV, float KiV, float KdV){
+void   Axis::setPIDValues(float KpPos, float KiPos, float KdPos, float propWeight, float KpV, float KiV, float KdV){
     /*
     
     Sets the positional PID values for the axis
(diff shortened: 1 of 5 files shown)


BarbourSmith:
The code itself looks beautiful and clean, but I'm not sure I fully understand what it does.

This let's us use both the PonM system and the regular system at the same time, but with weights on how much influence each have? That's really cool. Wouldn't a different Kp term be needed for each system?

BarbourSmith:
Sorry, should have read the linked article before asking. That's exactly what it does. Here's the relevant section from the author of the PID library:

> I didn’t add the following to the Arduino library code, but this is a feature that might be of some interest if you want to roll your own. Setpoint Weighting is, at it’s core, a way to have both PonE and PonM at the same time. By specifying a ratio between 0 and 1, you can have 100% PonM, 100% PonE (respectively) or some ratio in between. This can be helpful if you have a process that’s not perfectly integrating (like a reflow oven) and want to account for this.

krkeegan:
Yes, exactly.

In the simplest form, this gives the user direct access to enabling PonM by selecting 0.  But yes the ability to add weighting is interesting.  I haven't played enough with it to have an opinion, but allowing this doesn't seem to add much additional complexity for users.

PR #312, opened 2017-09-25
Fix Compiler Warnings that Appear with Default Settings

### LINKED ISSUE #290: Errors shown while compiling and uploading arduino code to arduino board
I followed the instructions to download Arduino IDE and download latest firmware (version 0.86 as I write this).  I opened the file "cnc_ctrl_v1.ino" and when I hit upload it appears to upload to the mega board but there are a long list of errors in orange that show up while it's compiling and getting ready to send it to the board.

I'm not deeply experienced with Arduino but is this normal every time you upload a file? It still says "done uploading" so I assume it's uploaded properly but I am not sure.

I have taken a screenshot which shows the Arduino IDE screen right after I finished uploading with all of the errors showing in orange.

Thanks,
Wade

maslow cnc arduino upload errors

krkeegan:
The default installation of the arduino compiler will display
some 'default' level of warnings.  This commit fixes those
errors so that no warnings are displayed by default.

There are additional warnings that can be seen selecting the
'All' setting, this does not fix all of those.

Fixes #290

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -1636,7 +1636,7 @@ int   findNextGM(const String& readString, const int& startingPoint){
     return nextGIndex;
 }
 
-void  interpretCommandString(const String& cmdString){
+void  interpretCommandString(String& cmdString){
     /*
     
     Splits a string into lines of gcode which begin with 'G' or 'M', executing each in order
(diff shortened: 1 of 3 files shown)


BarbourSmith:
Looks great to me!

I think some of these changes were made by @wrh3c 

@wrh3c do you feel OK about changing just these few lines? I like all of your changes from a best practices perspective, but we do get a good number of questions about the compiler warnings

wrh3c:
I reviewed the lines that are being changed, and I don't think I wrote any of them.  I made some changes in CNC_Functions.h, but not to the function declarations.

I'm all for fixing the code so we don't get compiler warnings.  That's a good practice!

PR #313, opened 2017-09-26
Indent Comments to Match Surrounding Code

krkeegan:
This is a bit mundane and self-centered, but the code folding in my IDE doesn't work correctly if the comments are not indented.

Hopefully everyone will tolerate my convenience.


BarbourSmith:
Love it

PR #314, opened 2017-09-28
Improvements to the Movement Calculations and PID Controllers

krkeegan:
**These changes will affect how your current PID tunings work, almost certainly causing them not to work, you will want to merge the corresponding pull request in GC as well, or manually make the changes to your PID tunings**

This makes a number of changes/improvements which include:

- Removing the running average on the input to the PID velocity calculation (this caused oscillations and damped the control reaction and I think averaging is an incorrect solution in control theory anyways)
- Adds a minimum rotation rate for the velocity controller (the motors are unable to turn at this rate, so this only caused integrator windup and noise)
- Ensure that position requests are equal to a whole number of encoder steps (prevents oscillations and integrator windup for unobtainable positions)
- Fix delay in the straight line loop in the movement loop to be consistent (ensures that the controllers are getting steps of equal length in each cycle)
- Fix a few other obscure design issues with the PID controller

Suggested PID Tunings that seem to work well:
Kp Pos = 1100
Ki Pos = 0
Kd Pos= 0
Proportional Weight 1

Kp Velo = 52
Ki Velo = 0
Kd Velo = 0

Corresponding Ground Control PR:
https://github.com/MaslowCNC/GroundControl/pull/430

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -60,7 +60,7 @@ void   Axis::loadPositionFromMemory(){
 
 void   Axis::initializePID(){
     _pidController.SetMode(AUTOMATIC);
-    _pidController.SetOutputLimits(-17, 17);
+    _pidController.SetOutputLimits(-20, 20);
     _pidController.SetSampleTime(10);
 }
 
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -34,7 +34,6 @@ motorGearboxEncoder(pwmPin, directionPin1, directionPin2, encoderPin1, encoderPi
     _pidController.setup(&_pidInput, &_pidOutput, &_pidSetpoint, 0, 0, 0, P_ON_E, REVERSE);
     
     //initialize variables
-    _direction    = FORWARD;
     _axisName     = axisName;
     _eepromAdr    = eepromAdr;
     
@@ -101,10 +100,6 @@ void   Axis::computePID(){
 
     if (_disableAxisForTesting || !motorGearboxEncoder.motor.attached()){
         return;
-    }
-    
-    if (_detectDirectionChange(_pidSetpoint)){ //this determines if the axis has changed direction of movement and flushes the accumulator in the PID if it has
-        _pidController.FlipIntegrator();
     }
     
     _pidInput      =  motorGearboxEncoder.encoder.read()/_encoderSteps;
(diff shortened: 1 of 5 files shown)


davidelang:
please make the minimum speed a variable defined up in the header (if not a 
configurable thing), different people are using dfferent motors because they 
couldn't get the kickstarter motors.

krkeegan:
I think I have tested all things I can think of at this point.

cnc_ctrl_v1/Axis.cpp
```diff
@@ -61,13 +60,18 @@ void   Axis::loadPositionFromMemory(){
...
     _pidController.SetMode(AUTOMATIC);
-    _pidController.SetOutputLimits(-17, 17);
+    _pidController.SetOutputLimits(-20, 20);
     _pidController.SetSampleTime(10);
 }
 
 void    Axis::write(const float& targetPosition){
     
-    _pidSetpoint   =  targetPosition/_mmPerRotation;
+    // Ensure that _pidSetpoint is equal to whole number of encoder steps
```
BarbourSmith: 👍

cnc_ctrl_v1/Axis.cpp
```diff
@@ -103,10 +107,6 @@ void   Axis::computePID(){
         return;
     }
     
-    if (_detectDirectionChange(_pidSetpoint)){ //this determines if the axis has changed direction of movement and flushes the accumulator in the PID if it has
-        _pidController.FlipIntegrator();
```
BarbourSmith: I'm glad to see the flip integrator leave. It was ugly and hacky. 👍 👍 👍

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -432,46 +432,50 @@ and G01 commands. The units at this point should all be in mm or mm per minute*/
     float aChainLength;
     float bChainLength;
     long   numberOfStepsTaken         =  0;
+    unsigned long beginingOfLastStep = millis() - delayTime;
     
     while(numberOfStepsTaken < finalNumberOfSteps){
+        //if enough time has passed to take the next step
```
BarbourSmith: ❤️

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -958,8 +963,7 @@ void  calibrateChainLengths(){
     
     //measure out the left chain
     Serial.println(F("Measuring out left chain"));
-    leftAxis.setPIDAggressiveness(.1);
```
BarbourSmith: Thanks for remembering to check on these values too

cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -958,8 +963,7 @@ void  calibrateChainLengths(){
...
-    singleAxisMove(&leftAxis, ORIGINCHAINLEN, 500);
+    singleAxisMove(&leftAxis, ORIGINCHAINLEN, 800);
```
BarbourSmith: Should we worry about the chain wrapping more at the faster feedrate of 800?

krkeegan: I tried both the chain calibration and machine measurement last night.  It doesn't seem unsafe, but maybe I have been working with it for so long.

BarbourSmith: I tested it too and it seemed very reasonable and the slight speed improvement is very welcome. I think I was just very conservative originally

davidelang:
one of the problems early on was the chain wrapping around the sprocket. you may 
have slowed it down to give people a better chance to deal with that without 
getting their fingers in the chain.

But it's far safer to just attach a string to the chain and use that to prevent 
the wrapup

PR #315, opened 2017-09-30
Enable 3D Linear Movements for G00 and G01

krkeegan:
Only works if zaxis is connected.  If not, zaxis movements remain as they
were.

Carefully calculates the feedrate to ensure that the limited zaxis
feedrate is not exceeded.

Only affects G00 and G01 commands

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -180,6 +180,13 @@ void   Axis::changePitch(const float& newPitch){
     _mmPerRotation = newPitch;
 }
 
+float  Axis::getPitch(){
+    /*
+    Returns the distance moved per-rotation for the axis.
+    */  
+    return _mmPerRotation;
+}
+
 void   Axis::changeEncoderResolution(const int& newResolution){
     /*
     Reassign the encoder resolution for the axis.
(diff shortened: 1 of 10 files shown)


blurfl:
Wouldn't this benefit from the fix for the hard-coded a-axis pitch?

krkeegan:
Done.

krkeegan:
I am a little concerned that some of the choppy behavior that I have been seeing when traversing at the top of the work surface is caused by the motor shield overheating.

I am seeing this choppy behavior occur when all three axes are attached even when not using this branch.  I am concerned that this branch may make the situation worse because it would cause all three axes to be attached for all linear moves.

I would appreciate it if some people would test this branch.  You don't need to have complicated 3D contours, even a simple square with multiple depth passes would likely be a good test.

blurfl:
I've run your Square 5.5 Forum.nc with this branch and did not see the choppy behavior you mentioned. There were some times when the horizontal movement paused waitinfg for the z-axis change tom complete, but I don't think that's what you're describing. It seems OK on my setup with GC0.92

BarbourSmith:
I vote for merging the branch as is, but keeping an eye on the situation. I have seen in the past changes to the PID controller affecting the thermal behavior of the drivers, but this branch also includes excellent changes that the community is asking for. Thank you @krkeegan for both the excellent contributions and for pointing out that we need to keep an eye on if we're stressing the driver chips more

PR #316, opened 2017-09-30
Fix Compiler Errors; Remove Unused Variables and Functions

krkeegan:
Compiler errors were unused variables or comparison of signed and
unsigned.

Also cleaned up numerous other unused variables and functions.

Saved 2% of dynamic memory.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -19,14 +19,11 @@
 #include "Arduino.h"
 #include "Axis.h"
 
-#define FORWARD 1
-#define BACKWARD -1
-
 #define EEPROMVALIDDATA 56
 #define SIZEOFFLOAT      4
 #define SIZEOFLINSEG    17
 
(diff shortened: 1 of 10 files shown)


cnc_ctrl_v1/Axis.cpp
```diff
@@ -19,14 +19,11 @@
...
 #include "Axis.h"
 
-#define FORWARD 1
-#define BACKWARD -1
-
 #define EEPROMVALIDDATA 56
 #define SIZEOFFLOAT      4
 #define SIZEOFLINSEG    17
 
-Axis::Axis(const int& pwmPin, const int& directionPin1, const int& directionPin2, const int& encoderPin1, const int& encoderPin2, const String& axisName, const int& eepromAdr)
```
BarbourSmith: ❤️

krkeegan:
Unused variables that are declared inside a function show up as warnings in the compiler.

Sadly globals do not.  I did a search for the global variables that seemed like they were unused and removed those that didn't seem like they would ever be used.

And some of the other fixes are just things I noticed while working with the code.

PR #317, opened 2017-10-03
Fix Movement of Z Axis for Non-Standard Pitches

krkeegan:
Including apparently the ability to set a negative pitch for zaxis adjustments that spin the opposite direction.


krkeegan:
This is included in #315, I made a separate request here in case we want to merge them separately.

PR #318, opened 2017-10-12
Improvements For Low RPM Speed Calculation

krkeegan:
- Adds an elapsedtime attribute to the encoder which returns the number of microseconds between the last two encoder steps, this allows for more accurate resolution of the RPM at very low speeds.
- Dampen some of the quantization that occurs when using steps to calculate RPM
- Change ComputeSpeed to private function since calling it outside of the PID calculation causes the speed to be calculated on less than a full loop.
- Add CachedSpeed function to obtain the RPM last returned by ComputeSpeed, useful at least for PIDVelocity tuning function.


BarbourSmith:
The idea of keeping track of the uS between pulses to get better resolution at low RPM is brilliant 👍 👍

PR #319, opened 2017-10-13
Configurable Loop Timing; Improved Movement Loop Timing; and Debugging

krkeegan:
Adds a compiler variable LOOPINTERVAL that sets the number of microseconds between each PID.  All places throughout the code that rely on the loop timing for any calculations are adjusted appropriately. 
 This allows an advanced user to test out a higher PID frequency.  

Also adds a compiler variable misloopDebug, that when set to 1 (defaults to 0) will warn the user whenever the movement loop fails to complete before being interrupted by the PID loop.  This is useful for testing and finding your fastest LOOPINTERVAL setting.

Please be careful increasing the PID frequency, I am concerned based on subjective testing that there may be a heat dissipation issue when used at high frequencies.

The above changes will not affect a regular compile of the Firmware and but require a developer to alter the values before compiling to cause any change from how this function now.

Finally, I have removed the concept of a timer in the movement loop entirely.  Instead, each loop waits until the PID loop processes the previously sent command before sending the next command.  This is much simpler and less error prone.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -58,7 +58,7 @@ void   Axis::loadPositionFromMemory(){
 void   Axis::initializePID(){
     _pidController.SetMode(AUTOMATIC);
     _pidController.SetOutputLimits(-20, 20);
-    _pidController.SetSampleTime(10);
+    _pidController.SetSampleTime(LOOPINTERVAL / 1000);
 }
 
 void    Axis::write(const float& targetPosition){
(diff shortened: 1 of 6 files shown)


cnc_ctrl_v1/Axis.cpp
```diff
@@ -34,7 +34,7 @@ motorGearboxEncoder(pwmPin, directionPin1, directionPin2, encoderPin1, encoderPi
     _axisName     = axisName;
     _eepromAdr    = eepromAdr;
     
-    initializePID();
+    initializePID(loopInterval);
```
BarbourSmith: 👍 elegant

PR #320, opened 2017-10-17
Fix Error in Quantization Buffer; Allow RPM to Approach 0

krkeegan:
Silly typo in quantization buffer prevented it from doing anything.

If amount of time since last seen encoder step is greater than time
between last encoder steps, then use the time last seen to calculate
speed.  This allows us to at least see speeds approaching 0.  With
elapsedtime only, RPMs would rarely approach 0.

### DIFF
--- a/cnc_ctrl_v1/Encoder.h
+++ b/cnc_ctrl_v1/Encoder.h
@@ -125,6 +125,17 @@ public:
 		interrupts();
 		return ret;
 	}
+	inline int32_t lastStepTime() {
+		if (interrupts_in_use < 2) {
+			noInterrupts();
+			update(&encoder);
+		} else {
+			noInterrupts();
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Good catch!

krkeegan:
The quantization buffer is helping a lot with the high frequency noise.

At a constant speed of about 11RPM the buffer dropped the standard deviation of the voltage from 11 to 6 and the actual RPM from .41 to .35.  This is with the incremental velocity setup.

I think there is a chance that the making the buffer even more aggressive, could further depress this noise without many drawbacks.  

My hope is that this allows us to use more aggressive controller settings with less vibration.

BarbourSmith:
BEAUTIFUL!!!!!

Issue #321, opened 2017-10-21
G00 command movement error on z axis for negative values


blurfl:
I noticed in the z-axis panel that the raise and lower commands were not symmetrical - issue a 1mm raise followed by a 1mm lower and the router does not return to the spot is started from. This is easiest see when watching the motor coupler. It is also possible to see in the z calibration step of the calibration sequence - the 'move -0.1mm' button does very little except buzz.
 Looking into GC, a G00 Zxx command is sent in these instances. I found that if I changed the G00 to G01, the z movement was correct. Further, the positive case (raise the bit) works correctly with G00 or G01. G00 Z-xx will move the motor by an incorrect amount in every instance.
 A workaround in the short term would be to change the 'lower' commands in GC to send G01, but the issue is in the firmware handling of G00 on the Z axis. 
 I've looked around in CNC_Functions.h in the G1() function but haven't found joy. It looks to me like the resulting call to singleAxisMove(&zAxis, zgoto, zfeedrate) is the same whether G00 or G01 is used.
 I've found that using the macros to issue the command strings is helpful.

these are symmetrical, change the 'G01 Z-0' to 'G00 Z-0' to see the issue occur
#move .1mm (converted to inches)
G20 G91 G01 Z-0.003937 G90
G20 G91 G01 Z0.003937 G90

#move .1mm 
G21 G91 G01 Z-0.1 G90
G21 G91 G01 Z0.1 G90

#move 15 degrees converted to mm (3.175mm / 24)
G21 G91 G01 Z-0.132292 G90
G21 G91 G01 Z0.132292 G90

blurfl:
I've found more information on this.
 PR#315 includes a change that adds the zaxis to the coordinatedMove() in CNC_Functions.h. The function G1() calls coordinatedMove(). CoordinatedMove() is called for all three axes, with feedrate 1000 (a hardcoded value...) for a rapid move - G0 gcode. This seems to be a problem for the z axis in the negative direction, haven't looked for why. In the positive direction the z axis seems to work correctly. Changing the hard-coded value to 500 makes the z axis work correctly in the negative direction in the coordinatedMove() function.
 @krkeegan, you've studied this section more deeply than I - can you think of a reason that the negtive direction of the z axis should behave this way? Should we consider adding a separate feed parameter to coordinatedMove() for the z-axis?
 A temporary workaround could check for non-zero z axis movement just before the call to coordinatedMove() and set the feedrate to 500 if found.

[...]

PR #322, opened 2017-10-22
Avoid divide by zero error in coordinated move()

blurfl:
Division of two floats can be truncated to zero if cast as a long. finalNmberOfSteps is used as a denominator in subsequent calculations.
This corrects issue #321.


cnc_ctrl_v1/CNC_Functions.h
```diff
@@ -446,7 +446,7 @@ int   cordinatedMove(const float& xEnd, const float& yEnd, const float& zEnd, fl
     //compute feed details
     MMPerMin = constrain(MMPerMin, 1, MAXFEED);   //constrain the maximum feedrate, 35ipm = 900 mmpm
     float  stepSizeMM           = computeStepSize(MMPerMin);
-    long   finalNumberOfSteps   = abs(distanceToMoveInMM/stepSizeMM);
```
BarbourSmith: 👍 👍

BarbourSmith:
Great catch! That's a tricky one. This is your second really good tricky bug in a row @blurfl . Awesome job
 👍 👍 👍

PR #325, opened 2017-11-10
Seems to auto detect PCB v1.2 

BarbourSmith:
Clean up the PCB version auto-detection and add support for board version 1.2

Needs to be tested thoroughly on boards 1.1 and 1.0 before merging


BarbourSmith:
Tested against hardware 1.0, 1.1, 1.2. No changes to 1.0 or 1.1 adds support for 1.2 and creates a new variable which contains the PCB version number

PR #328, opened 2017-11-18
Support zAxisAuto settings from GroundControl (B12 Y)

skiz:
Firmware update that allows setting the AUX1 to to be active high for supporting M codes based on `B12 Y`

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -31,6 +31,7 @@ libraries*/
 Servo myservo;  // create servo object to control a servo 
 
 bool zAxisAttached = false;
+bool zAxisAuto = false;
 
 #define FORWARD           1
 #define BACKWARD         -1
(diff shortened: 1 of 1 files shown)


blurfl:
How about moving the pin definition up to the top where the AUX pins are defined, so it's clear which pins have been committed? AUX4 is the z-axis depth probe presently. When moving the definition out of the function, maybe name spindleControlPin would be better?

skiz:
@blurfl AUX1 has already been committed, and I just made the existing code configurable.  AUX1 has 2 different usages depending on the configuration (servo signal vs 5v).  I'll clean this up a bit.

BarbourSmith:
@blurfl you have the most in depth knowledge of this section of code, thank you for taking a look at it. Does it look ready to merge to you?

@skiz Thank you for all of these improvements! What a cool thing to see first thing on a Monday morning 😀

blurfl:
It looks good. Thanks for cleaning up the AUX4 definition as well. Ship it!

PR #330, opened 2017-11-21
Halt on serial buffer overflow

blurfl:
RingBuffer.::write() will return a value to indicate the buffer overrun, so the logic in readSerialCommands() can trigger an immediate stop in the same way other portions of CNC_Functions.h do:
stopFlag = true;
checkForStopCommand();
This will have the same effect as using the ‘Stop’ button in GC, but won't need to go through the serial channel to take effect.
This seems like a minimal change that uses the existing methods and routines…

To test this, create a macro in GC with more than RingBuffer::BUFFERSIZE (currently 128) characters...


BarbourSmith:
Looks great! Awesome job I will merge this right after running the build

PR #331, opened 2017-11-22
Recompute half width and half height

blurfl:
recomputeGeometry fails to recompute halfWidth and halfHeight.


blurfl:
This could be part of some of the issues we've ben seeing. To prove the situation, print the value of kinematics.halfWidth before bedWidth is changed in CNC_Functions.h::updateKinematicsSettings() and after recomputeGeometry() is called at the end of the same function.

BarbourSmith:
Great catch @blurfl!!!

PR #332, opened 2017-11-29
Add support to the B02 function to extend only one or the other chain

BarbourSmith:
Which chain to extend is indicated with an L or R added to the line

Eg 'B02 L '

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -1054,30 +1054,32 @@ void  G38(const String& readString) {
   }
 }
 
-void  calibrateChainLengths(){
+void  calibrateChainLengths(String gcodeLine){
     /*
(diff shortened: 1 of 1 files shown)


krkeegan:
I don't know the proper best practices for gcode.  But is it common to just have a variable without a value trigger something?  Would L1 or R1 or Z1 be the better nomenclature?

BarbourSmith:
That's a good question to raise.

I'm not sure. I think it's more clear with just the letter. Adding a number would make me as the question what the number means. I agree that it breaks with gcode convention

krkeegan:
Just a thought, I don't think it matters enough to fight over it.  I came across the same problem in the PID B codes and opted to use a 1 to mean true.

BarbourSmith:
You know the more I think about it the more I agree. I think that the correct behavior is that the command should always be 'B02 L1 R0' to just do one side. That way we have the option of moving more than one axis with a single command. Thanks for the pointer @krkeegan I will update the behavior

Issue #188, opened 2017-04-28
3 Axis Coordinated movement for linear interpolation


mindeye:
Currently, just X and Y axis movements are coordinated. In some cases, having all 3 axis (X, Y, and Z) moving in a coordinated fashion would lend itself to doing more complicated 3D cuts. I'm envisioning this working in the manner that the cutter moves in a straight line in 3D space from it's start point to the end point specified by the X,Y,Z coordinates.

The gcode I saw in my output from Fusion 360 looked something like this (in terms out output, not values)

G01 X1.3 Y1.3 Z1.3

Let me know if any other info would be helpful.

BarbourSmith:
Looks perfect! Thanks for filing the issue 💯 👍

krkeegan:
This is fixed by #315

PR #333, opened 2017-11-30
Really Fix 3D Linear Movements

krkeegan:
So this is embarrassing.  When I submitted PR #315 I created the branch from a cherry-picked set of commits.  I guess I missed something because I just discovered that 3D movements haven't ever been working on the master branch.  Even though I had it working at some point.

Anyways, this PR actually solves #188 and enables 3D linear movements for G00 and G01.  I checked and it really works this time.


BarbourSmith:
Thanks for testing and fixing it! 😀

PR #334, opened 2017-11-30
Clicking Stop, Stops the Axis Where it is Not Where it Should Be

### LINKED ISSUE #293: Stop Command Doesn't Strictly Stop Motors
If the user presses the Stop command in GC, the motors stay active for two seconds before turning off.

This happens because the firmware doesn't treat STOP as a true e-stop.  Instead, it tries to hold the motors at the position they are supposed to be.  This action means the motors spend 2 seconds (in the motor hold function) trying to dial in this position as best as possible.  

In an ideal world, the motors are in the positions they are supposed to be and the two seconds is just filled with a lot of humming but no real movement.

However, if there is an error and the motors are not where they are supposed to be, the motors may move (and move a lot if the error is large) for up to two seconds.

Obviously, this can be a bit unnerving if you are near a crash, and this probably isn't the safest.

A stop should bring the motors to a complete stop as fast as possible.  I think this can be achieved by detaching the motors immediately rather than holding first.

krkeegan:
Fixes #293 

This adds a stop function to Axis.  Calling it stops the motor where it is, not where it should be.  This should ensure that even when things have gone wrong, there will not be a 2 second lag in stopping the axis.

Also unifies the checkforstopcommand function, so that its actions are consistent whenever a stop command is encountered.

PR #333 should be merged first, otherwise there will be an oddity in stopping the zaxis.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -242,6 +242,19 @@ void   Axis::endMove(const float& finalTarget){
     
 }
 
+void   Axis::stop(){
+    /*
+
+    Immediately stop and detach the axis, no matter where it is
+
+    */
+
+    _timeLastMoved = millis();
+    _axisTarget    = read();
+    detach();
+
+}
+
 void   Axis::wipeEEPROM(){
     /*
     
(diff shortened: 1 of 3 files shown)


BarbourSmith:
Fantastic!!

PR #336, opened 2017-12-01
Make firmware read B02 L1 R0

BarbourSmith:
Make the firmware look for values for the left and right axis on the B02 command

Thank you @krkeegan for pointing out that the old implementation of this was clunky. I appreciate you taking the time to look at my code and offer insightful suggestions like that. I like the new protocol better and I like the way it is implemented now much more. It's in line with the way all of the other commands are read.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -1031,7 +1031,7 @@ void  calibrateChainLengths(String gcodeLine){
     in length
     */
     
-    if (gcodeLine.indexOf('L') != -1){
+    if (extractGcodeValue(gcodeLine, 'L', 0)){
         //measure out the left chain
         Serial.println(F("Measuring out left chain"));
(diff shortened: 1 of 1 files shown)


krkeegan:
Awesome, the hive mind generally produces a good result.

krkeegan:
I have been reviewing the grbl code trying to learn how they do things and I came across this:

`// Import the next g-code word, expecting a letter followed by a value. Otherwise, error out.`

https://github.com/gnea/grbl/blob/5967839ab38f8e984a384db1f64e781cffb9e739/grbl/gcode.c#L119

I also found this NIST paper which also seems to be the most current authoritative standard for gcode??

http://ws680.nist.gov/publication/get_pdf.cfm?pub_id=823374

In section 3.3.2 it basically says that a word is a letter followed by a number, it doesn't allow for letters without a subsequent number.

BarbourSmith:
Fantastic! I'm glad to hear we're doing it right 😀 . Thanks again for pointing me in that direction

PR #337, opened 2017-12-09
Include Sprocket Geometry in Triangular Kinematics

reecej:
Updated the chain length calculations for triangular kinematics to include sprocket geometry effects.

Accounted for chain length wrapped around the sprocket as well as difference in distance of the chain on the sprocket compared to the motor axes.

Additional discussion, as well as an explanation of the geometry involved can be found at: https://forums.maslowcnc.com/t/accounting-for-motor-sprocket-induced-error-when-calculating-chain-lengths/1422

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -193,8 +193,13 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
     //Confirm that the coordinates are on the wood
     _verifyValidTarget(&xTarget, &yTarget);
     
-    float Chain1 = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
-    float Chain2 = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    //Calculate motor axes length to the bit
+    float Motor1Distance = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    float Motor2Distance = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    
+    //Calculate chain lengths accounting for sprocket geometry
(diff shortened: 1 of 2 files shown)


blurfl:
I wonder whether Motor1Distance and Motor2Distance would be better calculated in Kinematics::recomputeGeometry()? They only change when the calibration changes and keeping the inverseKiinematics routine free from as much calculation as possible will help with speed.

Too, I wonder whether those values and Chain1 and Chain2 might be useful to the simulator? How to make them available to GC?

reecej:
Motor1Distance and Motor2Distance are actually the exact same as the previous triangular kinematics formulae, and represent the distance from the motor axes to the sled bit. As such, they change every time xTarget or yTarget change.

The selected variable name may not be as descriptive as it could be.

reecej:
Corresponding changes have been made for the GC simulator as well: [PR #492](https://github.com/MaslowCNC/GroundControl/pull/492)

cnc_ctrl_v1/Kinematics.cpp
```diff
@@ -193,8 +193,13 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
...
     _verifyValidTarget(&xTarget, &yTarget);
     
-    float Chain1 = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
-    float Chain2 = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    //Calculate motor axes length to the bit
+    float Motor1Distance = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    float Motor2Distance = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
+    
+    //Calculate chain lengths accounting for sprocket geometry
+    float Chain1 = (R * (3.14159 - acos(R/Motor1Distance) - acos((_yCordOfMotor - yTarget)/Motor1Distance))) + sqrt(pow(Motor1Distance,2)-pow(R,2));
```
BarbourSmith: Very clean implementation! Nicely done

PR #339, opened 2017-12-11
Fix for z not retracting before moving home

BarbourSmith:
What's happening is that since checkForStopCommand() is not called unless the machine is moving. A stop stop sent when the machine is not running causes issues.

The stop command (!) will sit in the buffer until a move command is sent. As soon as the move command is processed, the stop is detected and the move command is not executed.

The solution is to also check for the stop command when the machine is not moving.


krkeegan:
good catch.

PR #341, opened 2017-12-12
Update CNC_Functions.h

madgrizzle:
Current code only pauses when the next tool ID is >0.   By comparing nextTool to lastTool, all changes in tools will result in a pause, including when changing back to T0.

This happened to me this weekend when the gcode makercam produced put my 1/8-bit as T1 and my 1/4-inch bit as T0.  I did my pocket and follow work with T1 and then cut out the piece with T0.  The firmware did not pause when it got the command to switch back to T0.


blurfl:
Do we need to set lastTool = -1 when it is defined in order to not lose the first usage? Haven't run the code, just thought about this.

BarbourSmith:
That looks great to me! I like it as is where it won't prompt for tool zero 👍 

Congratulations on your first PR @madgrizzle !! 🎉 🎆

PR #342, opened 2017-12-13
Split up CNC_Functions; Improve Include Inheritance

krkeegan:
This is a doozy of a merge.

It splits up the ~2000 line CNC_Functions.h file into several much smaller files.  CNC_Functions had a number of problems including:

- It had implementing code in the header
- It was massive
- Logically unrelated code was all stuffed in there
- It didn't permit a good flow of inheritance that allowed define statements to take effect across files.

I tried not to change too many things (yet) and just reorganize.  But I had to fix a few things along the way.  I combined a number of our global variables into a single sys struct to make them easier to find.  Nearly all inheritance is handled in a single Maslow.h file, @BarbourSmith  the version control define is also in Maslow.h.

I truly believe that this is a necessary step in order to start more important improvements to the code base.

I have tested this and I believe everything works, but there are a few things I don't use like spindle control that could have latent issues.

I don't want to be pushy.  But we should merge this in the next few days as manually fixing any further improvements to CNC_Functions into this is a bit tedious.

This does NOT require any changes to GC.


davidelang:
I would suggest that you keep an eye on the grbl code and the trapizodial 
planning output it produces so that you can slowly morph the maslow code to 
where it can plugin there.

Thank you so much for doing this sort of work

krkeegan:
Agree, that is my goal as well.  I have to start from the bottom up, so the serial line parser is on deck first, then slowly work our way towards acceleration planning.

davidelang:
Also, there is a fork of the maslow firmware that drives stepper motors instead 
of DC motors, it would be great to be able to integrate that work at some point.

BarbourSmith:
yes yes yes yes yes yes yes!!! This has been on my todo list since the dawn of time. We've had this massive ugly header file with functions in it for way too long. Thank you so much for taking care of it. 

I'm going to merge this right away, we've got a week before the next official release so we can track down any lingering weirdness. 

This is massive. Thank you @krkeegan !!!! ❤️ 👍 👍 🔥

PR #343, opened 2017-12-15
Change Encoder Steps to Float

reecej:
The encoderSteps and zEncoderSteps variables were changed from int to float, and the changeEncoderResolution function updated accordingly.

Full discussion at: https://forums.maslowcnc.com/t/what-is-correct-chain-pitch/1430/

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -184,7 +184,7 @@ float  Axis::getPitch(){
     return _mmPerRotation;
 }
 
-void   Axis::changeEncoderResolution(const int& newResolution){
+void   Axis::changeEncoderResolution(const float& newResolution){
     /*
     Reassign the encoder resolution for the axis.
     */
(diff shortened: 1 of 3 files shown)


reecej:
I see that changes were made to a few of the files I modified here recently and I did not merge soon enough. The CNC_Functions.h changes should now go to System.cpp.

krkeegan:
We should also remove the extraneous lines from Axis::write() as well.

krkeegan:
Also nice work @reecej, good find.  I haven't had a chance to test it this solves all of it

reecej:
@krkeegan, are you talking about removing the steps*2, round, and /2? If so, I can include that as well.

Also, if you could test at some point and see if it works, that would be very much appreciated. I don't have my Maslow kit yet so I can't validate that this fixes the error.

krkeegan:
I made PR #344 which fixes the conflict so that @BarbourSmith doesn't have to figure it out.

I still need to test that this solves all of the problems, but I am pretty sure it does.

Nice work!

BarbourSmith:
Great! I will close this pull request and wait on merging #344 until we have confirmed that it fixes the issue

PR #347, opened 2017-12-16
Create Realtime Loop; Perform System Init on Stop

krkeegan:
Created a Realtime function that is called frequently to act on realtime actions.  Put all realtime functions in here. This includes returning position and error, reading the serial buffer.  Remove all calls to these functions in the code.

Remove all calls to Delay in favor of maslowDelay which calls the realtime function.

On Stop, return from all functions back to base loop and do a reset similar to init.

Sanitize command string before exectuting. Removes comments and other characters and functions which we don't support. This code was taken heavily from GRBL.

Clean up some usage of passed variables where not necessary in returnpoz

In RingBuffer use the passed buffer and don't constantly create large 128 byte strings.

Create a number of precompile defines to make changing some things easier and to remove need for variables in some places.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -213,6 +213,8 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
 }
 
 void  Kinematics::forward(const float& chainALength, const float& chainBLength, float* xPos, float* yPos){
+  
+    Serial.println(F("[Forward Calculating Position]"));
     
     float xGuess = 0;
     float yGuess = 0;
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -27,6 +27,10 @@ Kinematics::Kinematics(){
 
 }
 
+void Kinematics::init(){
+    forward(leftAxis.read(), rightAxis.read(), &sys.xPosition, &sys.yPosition);
+}
+
 void Kinematics::_verifyValidTarget(float* xTarget,float* yTarget){
     //If the target point is beyond one of the edges of the board, the machine stops at the edge
 
@@ -235,14 +239,18 @@ void  Kinematics::forward(const float& chainALength, const float& chainBLength, 
         
         guessCount++;
 
-        //Prevent the connection from timing out
-        Serial.print(F("[PEk:"));
-        Serial.print(aChainError);
-        Serial.print(',');
-        Serial.print(bChainError);
-        Serial.print(',');
-        Serial.print('0');
-        Serial.println(F("]"));
+        #if defined (KINEMATICSDBG) && KINEMATICSDBG > 0 
+          Serial.print(F("[PEk:"));
+          Serial.print(aChainError);
+          Serial.print(',');
+          Serial.print(bChainError);
+          Serial.print(',');
+          Serial.print('0');
+          Serial.println(F("]"));
+        #endif
+
+        execSystemRealtime();
+        // No need for sys.stop check here
 
         //if we've converged on the point...or it's time to give up, exit the loop
         if((abs(aChainError) < .1 && abs(bChainError) < .1) or guessCount > maxNumberOfGuesses){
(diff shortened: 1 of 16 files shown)


krkeegan:
This continues the work restructuring things from the bottom up.  The realtime function doesn't do much right now, but is a crucial element.

Next up, importing the gcode block structure concept from grbl.  Which should finally let me touch the $ variables.

blurfl:
Axis::test() calls motorGearboxEncoder.motor.directWrite() 1000 times in each direction; shouldn't that call happen before the loop? motorGearboxEncoder.motor.directWrite() sets PWM going, doesn't  need to be called repeatedly, does it?

cnc_ctrl_v1/cnc_ctrl_v1.ino
```diff
@@ -30,21 +30,16 @@ void setup(){
...
+    readyCommandString.reserve(INCBUFFERLENGTH);           //Allocate memory so that this string doesn't fragment the heap as it grows and shrinks
+    gcodeLine.reserve(INCBUFFERLENGTH);
+    Timer1.initialize(LOOPINTERVAL);
+    Timer1.attachInterrupt(runsOnATimer);
     
     Serial.print(F("PCB v1."));
     Serial.print(getPCBVersion());
     Serial.println(F(" Detected"));
-    
+    Serial.println(F("Grbl v1.00"));  // Why GRBL?  Why not Maslow?
```
blurfl: There was a desire for grbl compatibility, so this hardware might run on front ends written for grbl.

krkeegan:
No directwrite doesn't need to be repeatedly called, but it doesn't do anything if it is, so I just left it.  The only call that I am replacing in that function is a call to delay.  The realtime function take a couple hundred microseconds at the most and often times only 10 to complete.  A call to delay just wastes clock time, so we repurpose that time to check for stop commands, otherwise there is no way to stop the test function.

cnc_ctrl_v1/Motion.cpp
```diff
@@ -67,6 +76,8 @@ void movementUpdate(){
   movementUpdated = true;
 }
 
+
+// why does this return anything
```
BarbourSmith: I think that there was a plan to use the return value to check if the move had completed or if it was unable to reach it's target...but that is long gone so it could be a void() at this point

cnc_ctrl_v1/Motion.cpp
```diff
@@ -128,13 +137,13 @@ int   coordinatedMove(const float& xEnd, const float& yEnd, const float& zEnd, f
         if (!movementUpdated) {
             //find the target point for this step
             // This section ~20us
-            whereXShouldBeAtThisStep +=  xStepSize;
-            whereYShouldBeAtThisStep +=  yStepSize;
+            sys.xPosition +=  xStepSize;
```
BarbourSmith: ❤️

cnc_ctrl_v1/RingBuffer.cpp
```diff
@@ -98,23 +96,18 @@ String RingBuffer::readLine(){
...
         }
     }
-    // if there are no full lines in the buffer, an empty string will be returned
-    
-    return lineToReturn;
-    
 }
 
-String RingBuffer::prettyReadLine(){
+void RingBuffer::prettyReadLine(String &lineToReturn){
```
BarbourSmith: 💯 🥇

cnc_ctrl_v1/cnc_ctrl_v1.ino
BarbourSmith: Universal gcode sender in particular won't start until it has received the command "Grbl v1.00". It's a hack 😒

cnc_ctrl_v1/Kinematics.cpp
```diff
@@ -235,14 +241,18 @@ void  Kinematics::forward(const float& chainALength, const float& chainBLength,
         
         guessCount++;
 
-        //Prevent the connection from timing out
```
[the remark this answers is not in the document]
krkeegan: Understood. I left the output in as a debugging option, I think it could still be helpful.  It will now just output it's position and error data on the defined schedule if this takes too long.

cnc_ctrl_v1/cnc_ctrl_v1.ino
krkeegan: Well that is just silly.  Grbl provides a nice "ready" and "ok" commands for this purpose.  Oh well.

PR #350, opened 2017-12-27
Save Settings in EEPROM; Optimize RAM Usage

### LINKED ISSUE #183: Settings should really be stored in the firmware
The machine settings should really be stored in the machine's firmware. The firmware is connected to the machine. That the settings are not stored in the firmware causes a couple issues.

If you unplug from the machine, leave Ground Control open, plug back in the settings are not sent to the machine again. This means that the machine switches to using the default settings which is confusing and wrong.

I don't want to fix this issue until after next week's release because I just don't want to mess with the way the settings work and this is going to mess with all kinds of things. We're just starting to get things worked out, and I want at least one good week before we start changing things again. That way we can point back to version .6X as the good one.

krkeegan:
Created the $ command from Grbl.  This has the following commands:

- $ - List possible commands
- $$ - List current settings and their values
- $n=value - set a setting to this value
- $K - Force recompute of kinematics (settings noted with a $K affect the kinematics)
- $RST=$ - Wipe settings from eeprom (not step data) (this command is not listed by $)
- $RST=# - Wipe all maslow values including step data
- $RST=* - Wipe entire EEPROM of device all 4kb

All values previously passed by the B03 and B13 commands are saved on the device, plus a few more other helpful values as well. (dlang sorry I removed the gear teeth and pitch before you comments on the forum. they can always be added back in the future if there really is a need)

The settings struct has version control, so we can add new settings to be saved, although a jump from one version to the next will wipe the user's settings (not step counts) requiring a resync of the settings data.  Keep in mind any setting we add takes up a corresponding amount of RAM.

The step counts are saved in a different struct, that also has version control.  Because this is a much smaller struct and losing step data requires a full recalibration this function allows for upgrading from one version to the next.  I doubt there will be many needs for this, but it is an option.  We actually save steps now rather than chain lengths, so we don't need any machine details to know where the encoders are.

This seemlessly handles the upgrade from the old chain length data saved in EEPROM to this data, so users shouldn't need to recalibrate.  I have tested this and it works.  I also tested a freshly wiped GC and firmware and went through the calibration process just fine.

To avoid interfering with the prior EEPROM data we start at address 300.  It wastes the lower numbers, but I can't think of a scenario in which we use all 4kb of EEPROM.

With all of this, we now use ~100 fewer bytes on startup in RAM.  Most of this was achieved by using pointers in the axes and motorgearbox settings that are passed from the global settings to these classes.

This update REQUIRES a matching update to GC.  And should be peformed with a change in version numbers.

See https://github.com/MaslowCNC/GroundControl/pull/522

Closes #183


krkeegan:
Oh, I left a bunch of commented Grbl code.  I intend to uncomment and enable this code in Maslow overtime as features are added.  I commit to making sure that this comment code is eventually used or deleted.  I am not a fan of leaving nonfunctioning confusing code in the codebase for very long.

krkeegan:
Oh crucial thing.

YOU CANNOT GO BACK FROM THIS BRANCH WITHOUT RE-CALIBRATING YOUR CHAIN LENGTHS.

There is no downgrade feature.  Sorry.

cnc_ctrl_v1/Axis.cpp
```diff
@@ -96,7 +89,7 @@ void   Axis::computePID(){
         return;
     }
     
-    _pidInput      =  motorGearboxEncoder.encoder.read()/_encoderSteps;
+    _pidInput      =  motorGearboxEncoder.encoder.read()/ *_encoderSteps;
```
BarbourSmith: These pointers save us space, but they also ensure that if the value is changed in one place it will be changed everywhere, right? That's an improvement on two fronts!

cnc_ctrl_v1/Axis.cpp
```diff
@@ -215,13 +202,12 @@ bool   Axis::attached(){
...
 }
 
-void   Axis::hold(){
-    int timeout   = 2000;
-    
-    if (millis() - _timeLastMoved < timeout){
-        //write(_axisTarget*_mmPerRotation);
-    }
-    else{
+void   Axis::detachIfIdle(){
```
BarbourSmith: This is such a better name! 👍 👍 👍 👍

cnc_ctrl_v1/GCode.cpp
```diff
@@ -745,11 +707,11 @@ void  G38(const String& readString) {
       float zgoto;
 
 
-      float currentZPos = zAxis.target();
+      float currentZPos = zAxis.read();
 
       zgoto      = sys.inchesToMMConversion * extractGcodeValue(readString, 'Z', currentZPos / sys.inchesToMMConversion);
       sys.feedrate   = sys.inchesToMMConversion * extractGcodeValue(readString, 'F', sys.feedrate / sys.inchesToMMConversion);
-      sys.feedrate = constrain(sys.feedrate, 1, MAXZROTMIN * abs(zAxis.getPitch()));
+      sys.feedrate = constrain(sys.feedrate, 1, sysSettings.maxZRPM * abs(zAxis.getPitch()));
```
BarbourSmith: I really like the way you are using the sys structure. It feels similar to the way things are done on the GC side and it feels much more intuitive than what we had before

cnc_ctrl_v1/Report.cpp
```diff
@@ -19,21 +19,190 @@ Copyright 2014-2017 Bar Smith*/
...
-          && (incSerialBuffer.numberOfLines() < MAXBUFFERLINES) ) {  
-        Serial.println(F("ok"));
+    if (status_code == 0) { // STATUS_OK
+      Serial.println(F("ok"));
+    } else {
+      Serial.print(F("error: "));
+      #ifdef REPORT_GUI_MODE
+        Serial.println(status_code);
+      #else
+        switch(status_code) {
```
BarbourSmith: I like this system where we can send back informative error codes

BarbourSmith:
> We actually save steps now rather than chain lengths, so we don't need any machine details to know where the encoders are.

Fantastic! I think this change makes a lot of sense.

> This seamlessly handles the upgrade from the old chain length data saved in EEPROM to this data, so users shouldn't need to recalibrate. I have tested this and it works. I also tested a freshly wiped GC and firmware and went through the calibration process just fine.

Thank you for taking the time to think about and test that!

I will merge this right away and bump the version numbers to 1.011 to ensure we stay in sync 👍 👍

davidelang:
On Wed, 27 Dec 2017, BarbourSmith wrote:

> I will merge this right away and bump the version numbers to 1.011 to ensure we stay in sync 👍 👍

make it 1.0.1 or 0.1.1, not 1.011

BarbourSmith:
1.01.1 would be ideal, but that would break the version comparison system. Unfortunately 1.011 doesn't work with the version comparison system either because only two digits after the decimal are sent from the Arduino. I'm just going to bump us to 1.02

davidelang:
On Wed, 27 Dec 2017, BarbourSmith wrote:

> 1.01.1 would be ideal, but that would break the version comparison system. 
> Unfortunately 1.011 doesn't work with the version comparison system either 
> because only two digits after the decimal are sent from the Arduino. I'm just 
> going to bump us to 1.02

let's fix the version comparison system in 1.02 then as well :-)

let's allow it to support multiple dots in the version, leading zero's in a 
field should be able to be left out.

This would also have allowed us to go from 0.99 to 0.100 without confusing 
things (but the changes were significant enough that 1.0 was reasonable)

cnc_ctrl_v1/Axis.cpp
krkeegan: Yes, you are correct.  I still left the function call to changeEncoderResolution() in case we want any hooks to run on a change in the steps.

krkeegan:
@BarbourSmith for fun, once you have run GC on this once, you can open the arduino ide and select Serial Monitor under the tools menu.  If you select the proper port and set the baud to 57600 you can directly control the maslow from a command line.  The position reports that I complained about in #353 will be apparent.

BarbourSmith:
Oh yeah, I know exactly what you are talking about. The terminal gets totally spammed with position reports 😬

krkeegan:
> let's allow it to support multiple dots in the version, leading zero's in a
> field should be able to be left out.

I only add the caveat that Grbl seems to only allow X.Xx nomenclature. 

https://github.com/gnea/grbl/wiki/Grbl-v1.1-Interface#welcome-message

We have hardcoded 1.00 into our welcome message, I don't know if we care.

davidelang:
the grbl version string is one thing (and it's hardcoded to 1.0 to satisfy 
broken gcode senders), but for the maslow version string that we are worrying 
about here.

PR #362, opened 2018-01-06
Add a Simple Fake Servo Option

### LINKED ISSUE #361: Add Fake Servo Option for Testing
It would be nice if there was an ability to test changes to GC and the Firmware on only an arduino. 

I think the biggest issue is accurately triggering the encoder steps.

I see two options:
1. Software based - add a precompile define that alters the code at the lowest level we can to trick it into thinking that the motors are turning at the proper speed.  I think this would have to be done in MotorGearboxEncoder::computeSpeed().

2. Hardware version - Trigger the encoder pins.  Is there some way to jumper a few pins on the arduino and trigger the encoders using other pins?  Alternatively we could have a second arduino with a simple sketch that looks at the pwm signal and outputs appropriate encoder signals.

I am leaning towards the software version unless we can figure out how to do the hardware with only a single arduino.

I am open to any other suggestions/ideas.

krkeegan:
This is very basic.

If FAKE_SERVO is defined, each PID cycle it will look at what the last requested PID rpm was and will update the encoder steps as though a motor operated at that speed for the amount of loopinterval. 

It adds up to 10% error to mimic servo noise.  It doesn't do a great job of testing the PID settings (this would be tough) and doesn't verify that the encoder library is working.  But other than that I think this tests the rest of the code in the Firmware.

It closes #361

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -84,6 +84,15 @@ void   Axis::setSteps(const long& steps){
 
 void   Axis::computePID(){
     
+    #ifdef FAKE_SERVO
+      static unsigned long lastPrint = millis();
+      if (motorGearboxEncoder.motor.attached()){
+        // Adds up to 10% error just to simulate servo noise
+        double rpm = (-1 * _pidOutput) * random(90, 110) / 100;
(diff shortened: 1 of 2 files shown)


BarbourSmith:
I think that is a fantastic implementation. 

It does a great job of being a low level as possible which means the amount of code changed is minimal and we get to test as much of the code as possible. All the ways of doing this that I thought of were much more complicated. Beautiful.

👍 👍 💯

PR #368, opened 2018-01-10
Catch Line Ending on Quick Commands; Remove Watchdog

### LINKED ISSUE #366: Buffer overflow when using manual z-axis

krkeegan:
Need to discard any line endings sent after a quick command
otherwise they get interpreted as a blank line and an OK is
sent.  I am not super thrilled with how I wrote this, but I
think it is reasonably compact and logical.

REMOVE WATCHDOG.  This was also causing an issue when the
machine paused itself waiting for a manual Z change.  If
the buffer was empty, it would send an OK requesting a new
line.  We just need to be careful now to make sure that all
OKs are properly sent.

This along with changes to GC closes #366

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -38,19 +38,29 @@ void readSerialCommands(){
     if this is a necessary character write to the incSerialBuffer otherwise discard
     it.
     */
+    
+    static bool quickCommandFlag = false;
+    
     if (Serial.available() > 0) {
         while (Serial.available() > 0) {
             char c = Serial.read();
             if (c == '!'){
                 sys.stop = true;
+                quickCommandFlag = true;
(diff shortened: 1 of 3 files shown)


krkeegan:
We should carefully test, deleting watchdog could have some real issues if we forgot an OK response anywhere.  I think I have been careful to include all of them, but you never know.

BarbourSmith:
OK this passes basic testing for me, how do we feel about merging it now since we are a full 8 days away from the next release and that way we will test it thoroughly as we go through the week?

krkeegan:
I am around, so I should be able to fix things if issues pop up.  I will try and watch the forum for issues that might be related.  So I vote yes.

BarbourSmith:
Fantastic!

PR #369, opened 2018-01-11
Chain Sag Correction for Triangular Kinematics

reecej:
This updates the triangular kinematics formulas to correct for chain sag. It utilizes a new parameter, chainSagCorrection, which is determined by a new GroundControl calibration process.

I associated parameter number 37 with the new chainSagCorrection parameter, as I believe that was the next unused parameter.

An associated GroundControl PR will be submitted shortly.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -203,11 +203,23 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
     //Calculate motor axes length to the bit
     float Motor1Distance = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
     float Motor2Distance = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
-    
-    //Calculate chain lengths accounting for sprocket geometry
-    float Chain1 = (R * (3.14159 - acos(R/Motor1Distance) - acos((_yCordOfMotor - yTarget)/Motor1Distance))) + sqrt(pow(Motor1Distance,2)-pow(R,2));
-    float Chain2 = (R * (3.14159 - acos(R/Motor2Distance) - acos((_yCordOfMotor - yTarget)/Motor2Distance))) + sqrt(pow(Motor2Distance,2)-pow(R,2));
-    
+
+    //Calculate the chain angles from horizontal
+    float Chain1Angle = 3.14159 - acos(R/Motor1Distance) - acos((_yCordOfMotor - yTarget)/Motor1Distance)
+    float Chain2Angle = 3.14159 - acos(R/Motor2Distance) - acos((_yCordOfMotor - yTarget)/Motor2Distance)
+
+    //Calculate the straight chain length from the sprocket to the bit
+    float Chain1Straight = sqrt(pow(Motor1Distance,2)-pow(R,2))
+    float Chain2Straight = sqrt(pow(Motor2Distance,2)-pow(R,2))
+
+    //Correct the straight chain lengths to account for chain sag
+    Chain1Straight *= (1 + ((sysSettings.chainSagCorrection / 1000000000000) * pow(cos(Chain1Angle),2) * pow(Chain1Straight,2) * pow((tan(Chain2Angle) * cos(Chain1Angle)) + sin(Chain1Angle),2)))
+    Chain2Straight *= (1 + ((sysSettings.chainSagCorrection / 1000000000000) * pow(cos(Chain2Angle),2) * pow(Chain2Straight,2) * pow((tan(Chain1Angle) * cos(Chain2Angle)) + sin(Chain2Angle),2)))
+
+    //Calculate total chain lengths accounting for sprocket geometry and chain sag
+    float Chain1 = (R * Chain1Angle) + Chain1Straight;
+    float Chain2 = (R * Chain2Angle) + Chain2Straight;
+
     //Subtract of the virtual length which is added to the chain by the rotation mechanism
     Chain1 = Chain1 - sysSettings.rotationDiskRadius;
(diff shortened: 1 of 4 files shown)


krkeegan:
I haven't kept up with this.  But this is only a change to the triangular kinematics?  Does it not apply or is to complicated to apply to quadrilateral?

reecej:
Currently this is just for triangular kinematics. It could be applied to quadrilateral kinematics a well, but would require a significant overhaul of the kinematics formulas as well as the calibration process. Being as I don't have my Maslow yet I wanted to start here for testing and validation that the algorithm works as expected.

cnc_ctrl_v1/Report.cpp
```diff
@@ -199,6 +200,7 @@ void reportMaslowSettings() {
     Serial.print(F(" (z axis Kp Velocity)\r\n$34=")); Serial.print(sysSettings.zKiV);
     Serial.print(F(" (z axis Ki Velocity)\r\n$35=")); Serial.print(sysSettings.zKdV);
     Serial.print(F(" (z axis Kd Velocity)\r\n$36=")); Serial.print(sysSettings.zPropWeightV);
+    Serial.print(F(" (chain sag correction value, $K)\r\n$37=")); Serial.print(sysSettings.chainSagCorrection);
```
krkeegan: There is a mistake here.  It is confusing, but the start of the line describes the prior setting.  So what you want is the following line " (zaxis Velocity..." to be the start of this line and the " (chain sag correction ..." to be the following line.

reecej: Thank you! I was hoping I would be able to borrow your eyes to validate my code. I'll fix this.

cnc_ctrl_v1/Settings.cpp
```diff
@@ -102,6 +102,7 @@ void settingsReset() {
     sysSettings.zKiV = 0.0;    // float zKiV;
     sysSettings.zKdV = 0.28;   // float zKdV;
     sysSettings.zPropWeightV = 1.0;    // float zPropWeightV;
+    sysSettings.chainSagCorrection = 0.0;  // float chainSagCorrection;
```
krkeegan: This also needs to be declared in the header file.

krkeegan:
There are a few nits that need to be corrected.  Also, we need to increment SETTINGSVERSION whenever we change the struct.  But otherwise I have no other concerns.

@BarbourSmith if we can wait until tonight, I can submit a PR for the other changes to the sysSettings struct and update the SETTINGSVERSION to 2.

@reecej one other question.  Other than geometry, what affects this chainSagCorrection?  The amount of weight on the sled?  Are there other changes a user may make (short of redesigning their frame) that we should be telling them to recalculate this?

reecej:
I believe I fixed the errors, but it wouldn't be the first time I've missed something. I really appreciate the feedback!

The primary factors that affect chainSagCorrection are sled weight, chain weight, and workspace lean angle. Altering any of those parameters would affect this value. Other small things could have an effect as well, such as a dust collection hose being connected, etc. So I would recommend users set the machine up just as it would be during normal cutting for the calibration process.

cnc_ctrl_v1/Settings.h
```diff
@@ -42,6 +42,7 @@ typedef struct {  // I think this is about ~128 bytes in size if I counted corre
   float sledCG;
   byte kinematicsType;
   float rotationDiskRadius;
+  float chainSagCorrection;
```
krkeegan: I thought that struct members were stored in the order declared.  Although it appears I may be wrong about that:

https://stackoverflow.com/questions/281045/do-class-struct-members-always-get-created-in-memory-in-the-order-they-were-decl

Still, at the risk of being pedantic, I think this should be down after zPropWeightV.  It doesn't appear to matter but it will make it a bit more logical for us in the future $37 coming after $36.

reecej: Sounds good to me!

BarbourSmith:
> @BarbourSmith if we can wait until tonight, I can submit a PR for the other changes to the sysSettings struct and update the SETTINGSVERSION to 2.

@krkeegan Yes I will wait until we are ready to merge both simultaneously and increment the version number. We will do tests of the new system today.

@reecej  👍 👍 🔥

cnc_ctrl_v1/Kinematics.cpp
```diff
@@ -203,11 +203,23 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
...
     float Motor1Distance = sqrt(pow((-1*_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
     float Motor2Distance = sqrt(pow((_xCordOfMotor - xTarget),2)+pow((_yCordOfMotor - yTarget),2));
-    
-    //Calculate chain lengths accounting for sprocket geometry
-    float Chain1 = (R * (3.14159 - acos(R/Motor1Distance) - acos((_yCordOfMotor - yTarget)/Motor1Distance))) + sqrt(pow(Motor1Distance,2)-pow(R,2));
-    float Chain2 = (R * (3.14159 - acos(R/Motor2Distance) - acos((_yCordOfMotor - yTarget)/Motor2Distance))) + sqrt(pow(Motor2Distance,2)-pow(R,2));
-    
+
+    //Calculate the chain angles from horizontal
+    float Chain1Angle = 3.14159 - acos(R/Motor1Distance) - acos((_yCordOfMotor - yTarget)/Motor1Distance)
```
BarbourSmith: These guys are missing the ';' at the end of the lines

cnc_ctrl_v1/Settings.h
```diff
@@ -20,8 +20,8 @@ Copyright 2014-2017 Bar Smith*/
 #ifndef settings_h
 #define settings_h
 
-#define SETTINGSVERSION 1      // The current version of settings, if this doesn't
-                               // match what is in EEPROM then settings on 
+#define SETTINGSVERSION 2      // The current version of settings, if this doesn't
```
BarbourSmith: 👍

PR #372, opened 2018-01-16
Move Startup ok to Setup

krkeegan:
Otherwise we send an extra ok whenever a stop or error occurs.


BarbourSmith:
Great work catching this before it became an issue!

PR #374, opened 2018-01-29
Revise pwm motor control to avoid TimerOne

blurfl:
On the Rev.1.1 boards the PWM for the left motor is controlled by pin12. This pin is controlled by timer1 which is also used by the TimerOne library for much of the important timing for the firmware. This PR avoids PWM on pin12 altogether by applying the PWM to a different motor control.
 This should make it easier to tune the PID settings, and allow changing the PWM frequencies for the three motors (right motor is on timer3, z motor timer4, left motor now on timer2) without affecting other parts of the software. This PR does not alter the frequencies however.
 I've bench tested it to see that the motors move in the expected manner, but haven't run any of the frame calibrations with it yet. I think I've preserved the proper workings of the speed and _lastspeed variables.


blurfl:
I'll redo this and try to bring less 'old baggage' with it.

BarbourSmith:
Woah, what a cool idea!

PR #375, opened 2018-01-29
Rework motor commands to avoid PWM on pin12

blurfl:
On the Rev.1.1 boards the PWM for the left motor is controlled by pin12. This pin is controlled by timer1 which is also used by the TimerOne library for much of the important timing for the firmware. This PR avoids PWM on pin12 altogether by applying the PWM to a different motor control.

This should make it easier to tune the PID settings, and allow changing the PWM frequencies for the three motors (right motor is on timer3, z motor timer4, left motor now on timer2) without affecting other parts of the software. This PR does not alter the frequencies however.

I've bench tested it to see that the motors move in the expected manner, but haven't run any of the frame calibrations with it yet. I think I've preserved the proper workings of the speed and _lastspeed variables.

### DIFF
--- a/cnc_ctrl_v1/Motor.cpp
+++ b/cnc_ctrl_v1/Motor.cpp
@@ -90,16 +90,16 @@ void Motor::write(int speed){
 
       if(_pwmPin == 12){
           if (forward){
-              analogWrite(_pin1 , abs(speed));
-              digitalWrite(_pin2, LOW);
-              digitalWrite(_pwmPin, HIGH);
+              analogWrite(_pin1 , speed);  // PWM on this pin to avoid using pin12 (timer1)
+              digitalWrite(_pin2, LOW);    // this pin sets direction with respect to _pin1 
+              digitalWrite(_pwmPin, HIGH); // this pin enables the outputs of the other two
           }
           else if (speed == 0){
           }
           else{
-              analogWrite(_pin1 , 255 - abs(speed));
-              digitalWrite(_pin2, HIGH);
-              digitalWrite(_pwmPin, HIGH);
+              analogWrite(_pin1 , 255 - speed); // PWM signal is inverted to cause reverse motion
+              digitalWrite(_pin2, HIGH);        // setting this pin HIGH causes _pin1 PWM to drive reverse
+              digitalWrite(_pwmPin, HIGH);      // enable the output
           }        
       } else {
           if (forward){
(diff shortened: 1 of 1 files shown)


cnc_ctrl_v1/Motor.cpp
```diff
@@ -127,24 +122,33 @@ void Motor::directWrite(int voltage){
...
-        digitalWrite(_pin1 , LOW);
-        digitalWrite(_pin2 , HIGH );
-    }
-    
     if(_pwmPin == 12){
-        voltage = abs(voltage);
-        voltage = map(voltage, 0, 255, 0, 1023);  //Scales 0-255 to 0-1023
-        Timer1.pwm(2, voltage);  //Special case for pin 12 due to timer blocking analogWrite()
+        if (voltage > 0){
+            analogWrite(_pin1 , abs(voltage));
```
BarbourSmith: Is it safe to do PWM on these pins? I makes sense that it would be, but all the examples for using the part did the PWM on the other pin

BarbourSmith:
This is very cool! I will test also.

blurfl:
It is, and it works a treat!

cnc_ctrl_v1/Motor.cpp
```diff
@@ -83,42 +83,37 @@ void Motor::write(int speed){
...
-        }
-        
+      speed = constrain(speed, -255, 255);
+      _lastSpeed = speed; //saves speed for use in additive write
+      bool forward = (speed > 0);
+      speed = abs(speed); //remove sign from input because direction is set by control pins on H-bridge   
+
+      if(_pwmPin == 12){
+          if (forward){
+              analogWrite(_pin1 , abs(speed));
```
krkeegan: speed is already in + terms here, no need for the extra abs() call.  The same is true in the additional calls below.

blurfl: " add a note that it applies to specific board versions." Where to put this note? Dows it belong in  Motor.cpp, or in the description of the PR?

krkeegan: You are right, I think your added comments are sufficient.

BarbourSmith:
I've been trying to do some reading on this to make sure it's safe and it appears that it is! 

This forum post has some good information (more in the comments section):  https://www.raspberrypi.org/forums/viewtopic.php?f=37&t=90243

From looking at the block diagram the IN and EN lines are not handled any differently in terms of the hardware they touch inside the chip

image

And this is backed up by the electrical characteristics being identical:

image

Here is a good summary of both fast and slow decay modes mentioned in the first forum post I linked:

https://www.allaboutcircuits.com/technical-articles/difference-slow-decay-mode-fast-decay-mode-h-bridge-dc-motor-applications/

Basically we might see slightly different behavior when using the IN pin vs the EN pin for PWM because in one case the motor is being actively braked between pulse while in the other case it's allowed to coast, but it's totally safe to do and much better than the alternative of using different PWM frequencies which we KNOW has been giving people trouble which just effects one side or the other.

This is a brilliant plan @blurfl 👍 👍 Awesome.

krkeegan:
@BarbourSmith thank you for tracking down that data.

The "active braking" gives a little concern that the correct PID values for each main motor may not be identical.  But the PID controller may be robust enough to adjust to this difference and it sounds like your real world testing suggests this.

If I have a chance I will test this too.

blurfl:
I've played a bit trying to see 'active braking' but don't really see the effect.

 I've got motors on the  bench without any load, so it really isn't a true test, but when I run the 'Test Motors' command, the unloaded motors coast after the command finishes. The command comes from Axis.cpp, uses motor.directWrite() so PID is not involved, and finishes with a directWrite(0) which is special-cased in the motor.directWrite() code and makes a good place to try forcing the pins to specific states. I added a 'maslowDelay(2000);' between the direction commands in Axis.cpp to give the coast time to play out in the first direction. I see coasting on all three motors regardless of the state that I set the three pins.

BarbourSmith:
This tests great for me!

Is there any reason we should hold off on merging until after the release tomorrow? My feeling is that 1.04 doesn't have that many big new changes over 1.03 so I vote to merge now.

krkeegan:
i don't see any reason not too.  It seems like the real-world testing matches the theory.

BarbourSmith:
I can't think of any other tests. I poked it around a little bit more and it still seems great

BarbourSmith:
What a great idea @blurfl !! 👍 👍 Way to think outside the box!

PR #377, opened 2018-02-01
Changed rotationalDiskRadius to 250 from 100

madgrizzle:
Changed rotationalDiskRadius to 250 from 100 to match that in GC PR


BarbourSmith:
Great! 👍 👍

Issue #379, opened 2018-02-01
Machine won't start again after pressing stop


BarbourSmith:
Moved from GC issue [#603](https://github.com/MaslowCNC/GroundControl/issues/603)

For some reason my machine has stopped starting again after the stop button is pressed. It will stop immediately, but not move again until GC is restarted. 

This is a new behavior.

BarbourSmith:
Or maybe it's that we're getting one 'OK' instead of two.

From the log:

`Sent: !
ok
`

But when we print the amount of space GC thinks is available in the buffer it looks like this:

![image](https://user-images.githubusercontent.com/9359447/35699859-2a83ab38-0746-11e8-84b4-870726b47e22.png)

So there's `236 `spaces in the buffer because the line of gcode hasn't been processed yet, then there are `234` spaces because we've sent the "`!\n`" command, then we go back to `236` when it is marked that the stop command has been processed...so do we need to send an extra "OK" after stops are processed to indicate that the extra line has been cleared from the buffer? It has been cleared.

BarbourSmith:
I think this is related to #366

BarbourSmith:
Here's what I'm seeing is happening. 

The '!\n' is sent, and we immediately get an 'OK' status to indicate that it has been received from GCode.cpp line 51. Within the `coordinatedMove` function in Motion.cpp line 163 `if (sys.stop){return 1;}` is triggered and the machine stops right away (yay 🎉 ). 

The issue comes up because that return kicks us back out GCode.cpp line 663 where coordinated move was called, and nothing happens there to indicate that the line of gcode is complete `coordinatedMove(xgoto, ygoto, zgoto, 1000);.`

We then drop back up another to line 339 `G1(gcodeLine, gNumber);` and eventually kick out to line #594 which is `if (!sys.stop){reportStatusMessage(status);}` so we don't send the OK message there.

As far as I can tell !sys.stop is never cleared, right?

Is it safe to report the 'OK' after the stop command has been processed? I think it is because we've stopped sending gcode on the Ground Control side at that point.

I'm a little out of my element here @krkeegan do you have thoughts on what the best solution is?

krkeegan:
if sys.stop is true it causes the while loop in the main loop in cnc_ctrl_v1.ino at line 86 to bail out.  At which point we reset everything and clear the sys.stop flag.

@bar is this all on the current master versions of the firmware and GC?  And this is happening 100% of the time or is there some sequence I should try to reproduce?

[...]

PR #380, opened 2018-02-01
Fix for issue #379 (won't start again after stop)

BarbourSmith:
Fixes issue #379, the machine won't start again after stop. Basically, the line is being cleared from the buffer when the stop command is recieved, but no OK is being sent to indicate the space is free. It seems counter intuitive to send an OK message when the machine is stopped, but the buffer space is available.

Is this the right fix?

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -591,7 +591,7 @@ void gcodeExecuteLoop(){
       readyCommandString = "";
 
       // Get next line of GCode
-      if (!sys.stop){reportStatusMessage(status);}
+      reportStatusMessage(status);
   }
 }
 
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -233,36 +233,6 @@ void systemSaveAxesPosition(){
     }
 }
 
-// This should be the ultimate fallback, it would be best if we didn't even need 
-// something like this at all
-// TODO delete this function, we are not even using it anymore
-void  _watchDog(){
-    /*
-    If:
-      No incoming serial in 5 seconds 
-      Motors are detached
-      Nothing in Serial buffer
-      Watchdog has not run in 5 seconds
-    Then:
-      Send an ok message
-    
-    This fixes the issue where the machine is ready, but Ground Control doesn't know the machine is ready and the system locks up.
-    */
-    static unsigned long lastRan = millis();
-    
-    if ( millis() - sys.lastSerialRcvd > 5000 &&
-        (millis() - lastRan) > 5000 && 
-        !leftAxis.attached() and !rightAxis.attached() and !zAxis.attached() &&
-        incSerialBuffer.length() == 0
-       ){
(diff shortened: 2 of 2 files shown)


blurfl:
That certainly seems to fix it. Quick work!

BarbourSmith:
@krkeegan I don't feel like this is the right way either. I won't touch it until you have a chance to give it a look

krkeegan:
So while I think this solution is wrong, it sounds like an extra ok will solve the issue.  hmm why is that?  

Is there an additional command being placed in GC queue, or something funky about the character buffer counter in the GC queue?

krkeegan:
Plus, why is this cropping up now?  If you have the time and inclination, can someone roll back to v1.04 or v1.03 and see where this goes away?

BarbourSmith:
I started exploring on the GC side and here's what I found:

When we print the amount of space GC thinks is available in the buffer it looks like this:

image

So there's `236 `spaces in the buffer because the line of gcode hasn't been processed yet, then there are `234` spaces because we've sent the "`!\n`" command, then we go back to `236` when it is marked that the stop command has been processed...so do we need to send an extra "OK" after stops are processed to indicate that the extra line has been cleared from the buffer? It has been cleared.

BarbourSmith:
Help me understand why we do not want to send the OK command after a line of gcode has been executed, but not completed because a `!` command was sent?

blurfl:
Firmware1.02 does not show the issue, Firmware1.03 does.

krkeegan:
oooh ooh, I think I understand what you are describing now.

So let me see if I have this right.  

1. Send some command "G0 X5"
2. While the command is executing send "!"
3. Firmware sends "ok" response to "!" command
3b. GC sees the "ok" recognizes that it was in response to a quick command and clears only that portion of the buffer.
4. Firmware bails out of routines that it is running
5. Firmware dumps contents of serial buffer
6. Firmware waits for new commands
7. Meanwhile GC is still waiting for confirmation of "G0 X5" command
8. aaaaand we are stuck

I think I get it.

I see two solutions:
1. The firmware sends the necessary "ok"s to clear everything
2. GC recognizes that a stop dumps the buffer contents on the machine and clears it's own tracking buffer when it receives the stop "ok"

I think this is all a matter of perspective and what we want an "ok" message to mean.

I am inclined to say GC should handle this, that the firmware is correct in not sending an ok to the "G0 X5" command because that command never successfully completed.  Plus this design would allow us to queue more than one line on the machine (which I think we can do eventually) otherwise if we had multiple lines, we would need the firmware to say "ok" to those extra lines in the buffer, even though it hadn't even started performing them.

Thoughts? I am open to considering the alternative.

blurfl:
I like your GC approach.

BarbourSmith:
> Plus this design would allow us to queue more than one line on the machine (which I think we can do eventually) otherwise if we had multiple lines, we would need the firmware to say "ok" to those extra lines in the buffer, even though it hadn't even started performing them.

This is the deciding factor for me, I vote to handle this on the GC side. Too. Well summarized.

PR #383, opened 2018-02-04
Fix bug in storing encoder steps

reecej:
Fix bug where encoder steps are overwritten each time the board is booted up. This is related to the chain on bottom functionality, although the bug occurs whether the chain is on top or bottom.

### DIFF
--- a/cnc_ctrl_v1/Settings.cpp
+++ b/cnc_ctrl_v1/Settings.cpp
@@ -22,12 +22,6 @@ Copyright 2014-2017 Bar Smith*/
 
 #include "Maslow.h"
 
-void settingsInit(){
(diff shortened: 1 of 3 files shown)


krkeegan:
I gather that it wasn't working to call setupAxis() after loading the steps from eeprom?

reecej:
Correct. When setupAxes() was run, it would read the current encoder position and set the steps to that value, really just to initialize the values. But that was overwriting our loaded previous steps value, and hence the chains were having to be recalibrated at every system boot.

BarbourSmith:
This works great for me as well. No need to do the chain re-cal 👍 👍 Excellent job tracking down what the issue was!

PR #384, opened 2018-02-06
Forward kinematic improvements; Sync Settings

krkeegan:
This enables the sync settings function from https://github.com/MaslowCNC/GroundControl/pull/608

It removes the concept of $K.  Now anytime a setting which alters the kinematics calculation is changed, the forward position is recalculated.

Previously forward calculations could take up to 6 seconds, this change limits the search space to no more than the length of the chain, which can be set in ground control.  This results in a maximum search of about .6 seconds.

This requires a simultaneous update to GC, so the versions numbers should be changed.


BarbourSmith:
@krkeegan you just keep making these amazing, well thought out, well executed PRs. I never know what to expect in the morning 👍 👍 

I will test this right away (although I am sure it works great).

In the future whenever you want the version numbers bumped in a PR feel free to include it in the PR. I'm not too worried about skipping numbers in releases

BarbourSmith:
No rush, my computer is installing updates right now so I can't test or merge anything for a little bit anyway.

Excellent testing @blurfl 👍 👍

krkeegan:
ahh, found it.  Now just need to figure out the solution.

krkeegan:
ugh, more yak shearing, now it seems that the FAKE_SERVO option doesn't work anymore?

krkeegan:
I think this is working now.

I have found a reproducible bug, at least when using Fake_Servo(I doubt that matters).  The process is:

1. Load up GC.
2. Manually move the sled up
3. Stop before the move is complete
4. Somewhat quickly (not sure if this matters) open up advanced settings
5. Enable or disable custom PID position settings
6. GC will then send about 4 of the 6 commands and get stuck
7. No further commands can be sent

This seems like an ok/bufferspace issue.  I really wish these things would die.  At first glance I don't see any problems in the log.  So I have to add some debugging to GC to see if the bufferspace is the issue.

krkeegan:
good news is that clicking stop gets us out of the locked GC state, so that is at least a good rescue plan.

krkeegan:
OK, everything should work now.

BarbourSmith:
I'll give it a test too, but the gist is that we should or should not merge because of the issue in #387

krkeegan:
Yes, I that is my opinion.  I am having a hard time figuring out why this is happening.  Give it a test and see if you see the same issues with pressing stop.

BarbourSmith:
Great! I'm just getting my computer back online now and I should be able to test in the next hour or so...what a strange bug 🐛

blurfl:
Following your excellent series of steps to cause the issue, I see it on a board with motors attached. It stops after sending 5 or six lines and unfreezes after using the Stop button again.

krkeegan:
Uhoh, @blurfl you are seeing the sled position change as a result of multiple stop button presses?

blurfl:
I followed [these steps](https://github.com/MaslowCNC/Firmware/pull/384#issuecomment-363602846), haven't tried the sled move sequence yet. I'm right at the console, what steps to make it happen?

krkeegan:
ooh, sorry, to many commingled issues.  That one you linked to is fixed.  I found the issue and made the change in GC.  #387 is the thing I am still checking on.

blurfl:
So loading the firmware from this PR, and the GC version from PR #608, yes?
 I had a version of your 'Settings Improvements' branch from a couple days ago, just loaded it and ran it against the appropriate firmware. I did see the 'position loaded at' values creeping up by small amounts, though the motors did not run (changes too small).
 I went and got your most recent version from that branch and that fixes the issue. Using that branch and the firmware from PR #384 the 'position loaded at' values do not change and the motors do not run.

blurfl:
I confirmed that the z-axis doesn't preserve the units, opened a separate issue for that.

krkeegan:
Awesome, yes I confirm the same. The issue doesn't happen With a functioning machine.

PR #385, opened 2018-02-06
Change PWM signalling to use INx pins for PWM

blurfl:
Using the INx pins (_pin1, _pin2) directly for PWM will make this firmware compatible with other motor driver chips that don't implement the EN enable signal ('_pwmPin').


BarbourSmith:
Very cool! So does this mean we don't need the enable line at all anymore?

blurfl:
On the present bridge chip, we need it. Other chips (TLE5206) would show errors on that pin: over-temp and over-current.
 The L298 is designed to use the EN pin along with an some analog circuitry monitoring the sense output to implement over-current protection. We're not using it that way.

BarbourSmith:
First of all I want to apologize for taking so long to test this. I have really been struggling with my computer this last week (getting windows 10 to update on a 32gb disk is a mess)...I think I am all sorted out now, which is great.

When I run this firmware, my machine does not move at all. If I switch back to the master firmware it goes again. What am I doing wrong?

blurfl:
There's no hurry to get to these, you're deep into frame redesign, Windows updates and shipment preparation 😀
 I'm sorry, much has changed in the Firmware master and I hadn't updated this branch to keep up. Exciting times... I've brought it current now. I only have a 'V1.1' board to test agains, but that shouldn't matter.

BarbourSmith:
Great! I like that the issue WAS my slowness 😀 

I'm testing on 1.1 also, I'll run it right now.

BarbourSmith:
Works like a charm! 🎉 👍 👍 

Sorry again for the delay 🙁

PR #386, opened 2018-02-06
Avoid using timer0 or timer1 for pwm

blurfl:
This builds on PR #385.
Timer0 is used for Arduino internal timing, timer1 is used by the TimerOne library for the encoder and PID portions of the firmware. By avoiding those we can alter the PWM frequencies and maybe improve performance.
Further, create a function to set the PWM frequencies of timers 2-4. Don't change it yet, different frequencies will require some PID tuning.


BarbourSmith:
Awesome!! 👍 👍 ❤️ 

I read somewhere a while ago that we should really be using ultrasonic PWM frequencies for motors of our size

blurfl:
I've played with 4096kHz, it wants some PID tuning.

krkeegan:
First a caveat, I know very little about any of this.

I could be wrong, but if it needs a change in the PID tuning, something else may be going on.  The only way I can see this causing this is if at 4000kHz the "voltage" produced by the PWM signal starts to drift in someway higher or lower than expected or the timers in the arduino start to drift see:

https://playground.arduino.cc/Main/TimerPWMCheatsheet

krkeegan:
Also, is that a typo do we mean 4096Hz not kHz.

blurfl:
You're right about the frequency, ~4000Hz is the step above 490Hz, and ~31000Hz is the step above that. I was looking at the slight lag of the red reticule behind the black one, but comparing to 490Hz it looks the same. The motors sure purr at 31000Hz!

Setting 	   Divisor 	Frequency
0x01 	 	1  		31372.55
0x02 	 	8 	 	3921.16
0x03  		32  		980.39
0x04 	 	64 	 	490.20   <--DEFAULT
0x05 	 	128  	245.10
0x06  		256  	122.55
0x07 	 	1024  	30.64

cnc_ctrl_v1/Motor.cpp
```diff
@@ -87,28 +87,31 @@ void Motor::write(int speed){
...
+              } else {
+                analogWrite(_pin1 , speed);
                 digitalWrite(_pin2 , LOW );
+              }
+            } else {
+              digitalWrite(_pin1 , LOW );
+              digitalWrite(_pin2 , LOW );
             }
-        } else {
-            if (_pin1 != 11) {          // avoid PWM using timer1
```
BarbourSmith: I am so glad we are getting rid of this special case if statement. It was ugly and your solution is so much cleaner 👍

BarbourSmith:
This is so cool!!! 👍 👍 👍 👍

PR #389, opened 2018-02-10
Add PWM frequency to system settings

blurfl:
This builds upon PR #385 and #386 to allow changing the PWM prescaler frequency from GroundControl.

This needs a cooresponding GroundControl PR (GroundControl #619).

 - add a system setting $39 fPWM
 - set the default value to 3 = 490Hz
 - change the prescaler for timers2-4 on change of fPWM
 - report the change if verboseDebug > 0

### DIFF
--- a/cnc_ctrl_v1/Report.cpp
+++ b/cnc_ctrl_v1/Report.cpp
@@ -166,6 +166,7 @@ void reportMaslowSettings() {
     Serial.print(F("$36=")); Serial.println(sysSettings.zPropWeightV, 8);
     Serial.print(F("$37=")); Serial.println(sysSettings.chainSagCorrection, 8);
     Serial.print(F("$38=")); Serial.println(sysSettings.chainOverSprocket);
+    Serial.print(F("$39=")); Serial.println(sysSettings.fPWM);
   #else
     Serial.print(F("$0=")); Serial.print(sysSettings.machineWidth);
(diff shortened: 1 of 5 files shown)


krkeegan:
You also need to increment the SettingsVersion number in settings.h as well.  This ensures that settings on the arduino are wiped and reloaded with the default and then loaded from GC.  Otherwise, the new variable or potentially other variables can have some very odd values or even overflows.

blurfl:
Interesting, you can change PWM frequency within a cut file using $39=  commands. 
$39=1 -> 31,000Hz    $39=2 -> 4,100Hz    $39=3 -> 490Hz
Most changes seem pretty benign, though changing fPWM into 4,100Hz causes quite a lag sometimes. Probably not something to be used in every-day cutting, but interesting to use while exploring 😀.

blurfl:
it seems like the lag can happen with  any speed change during a cut, but putting in a manual z-axis change (or a several-second long auto one) seems to give the affected routine (PID?) time to clear or catch up. Not that a speed change would be useful, but it does cast interesting shadows of some internal process.

blurfl:
To see it change, turn on 'verboseDebug' in Config.h...

BarbourSmith:
This is so mind blowing. Wow. I want to keep playing around with it more, but even just changing the settings and hearing the tone change is incredible. It's funny that the noise has been with us for so long and now we can change it. What a fundamentally cool thing! It feels like this PR really touches the heart of the machine. Awesome work @blurfl!! You are really on fire here.

Issue #378, opened 2018-02-01
Report unhookedup axis as an error if told to move


thormj:
It seems like I keep forgetting to re-attach the z-axis... I think if an axis is supposed to be moving, but the encoder hasn't changed after a time, some actions should be taken:
1. Stop all motion (temporarily?.... so you can resume after the fault is cleared)
2. Clear the PID loops for all axes (at least this one -- I wand the power to be off when I plug it in)
3. Send up a message to GroundControl via the same mechanism as the Manual Z change (so GC puts up a dialog box and waits for the user to press OK)
4. When GC says it's done, resume motion and try to continue.

With this patch, I think it would also be good to have a flag for "Air Cutting" (accept, but don't do anything with Z) if we don't already have one.

BarbourSmith:
I think this is a great idea. 

How will you detect that the motor is not plugged in?

I would recommend just triggering the built in pause function because we've already got good infrastructure around it

PR #393, opened 2018-02-18
Initial version of a stalled/unhooked axis detector

thormj:
For issue #378, but **this needs testing** (I wrote it on a plane flight and I won't be back to carving until Mar 6).

Algorithm (and if this is not clear from the patch, let me know and I'll throw in some more comments):
1. When calculating the position PID for each axis:  if it moved since last time, increment `moved`, if not, increment `stalled`.
2. If an axis is detached, clear both variables.
3. When the report runs, examine all axes; if it has stalled for >5 cycles (so 500ms), prepare a message:
3b. If it's moved before, say it's stalled.  If it hasn't moved before, say it's UNHK'd
4. If I have a message from step 3, `pause()` the machine (which hopefully detaches the axes which will clear the error)

Not sure about step 4 happening fast enough... maybe I need to put a "if not paused" around the block in `report.cpp`?

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -105,27 +105,30 @@ void   Axis::computePID(){
     if (_pidController.Compute()){
         // Only write output if the PID calculation was performed
         motorGearboxEncoder.write(_pidOutput);
-		
-		if (millis() - stallmillis > STALLTIMEOUT)
-		{
-			stallmillis = millis();
-			
-			long delta=lastcoder - motorGearboxEncoder.encoder.read(); 
-			
-			if(lastcoder == 0 || delta >  STALLSTEPS || delta < STALLSTEPS)
-			{
-				moved=lastcoder !=0;
-				lastcoder = motorGearboxEncoder.encoder.read();
-			}
-			else
-			{
-				stalled=1;
-				stalldelta= delta;
-			}
-		}
-    } else {
-      lastcoder=0;  //We've been deactivated; reset counts
-    }
+
+        if(_pidOutput) {
+            if (millis() - stallmillis > STALLTIMEOUT)
+            {
+                stallmillis = millis();
+                
+                long delta=lastcoder - motorGearboxEncoder.encoder.read(); 
+                
+                if(lastcoder == 0 || delta >  STALLSTEPS || delta < STALLSTEPS)
+                {
+                    moved=lastcoder !=0;
+                    lastcoder = motorGearboxEncoder.encoder.read();
+                }
+                else
+                {
+                    stalled=1;
+                    stalldelta= delta;
+                }
+            }
+        } else {
+            lastcoder=0;  //No motion requested; reset counts
+            stalled=moved=stalldelta=0;
+        }
+    } 
     
     motorGearboxEncoder.computePID();
 }
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -83,8 +83,8 @@ void   Axis::setSteps(const long& steps){
 }
 
 void   Axis::computePID(){
-	static unsigned long stallmillis;
-	static long lastcoder;
+	static unsigned long stallmillis=0;
+	static unsigned long lastencodervalue=0;
     
     #ifdef FAKE_SERVO
       static unsigned long lastPrint = millis();
@@ -111,22 +111,23 @@ void   Axis::computePID(){
             {
                 stallmillis = millis();
                 
-                long delta=lastcoder - motorGearboxEncoder.encoder.read(); 
+                long delta=lastencodervalue - motorGearboxEncoder.encoder.read(); 
                 
-                if(lastcoder == 0 || delta >  STALLSTEPS || delta < STALLSTEPS)
-                {
-                    moved=lastcoder !=0;
-                    lastcoder = motorGearboxEncoder.encoder.read();
+                if(lastencodervalue == 0 || delta >  STALLSTEPS || delta < -STALLSTEPS)
+                {   
+                    moved=lastencodervalue !=0;                             //If not the first time through the function and we gots a lot of delta, we've moved.
+                    lastencodervalue = motorGearboxEncoder.encoder.read();  //Keep track of lastencodervalue; this inits it for the next round.
                 }
                 else
                 {
-                    stalled=1;
-                    stalldelta= delta;
+                    stalled+=1;
+                    if(stalled>240) stalled=240;    //No wrapping!
+                    stalldelta= delta;              //Debug: Keep track of the delta; if it shows up as 8 all the time, we should change STALLSTEPS.
                 }
             }
         } else {
-            lastcoder=0;  //No motion requested; reset counts
-            stalled=moved=stalldelta=0;
+            lastencodervalue=0;     //No motion requested; reset counts
+            stalled=stalldelta=0;
         }
     } 
     
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -83,6 +83,8 @@ void   Axis::setSteps(const long& steps){
 }
 
 void   Axis::computePID(){
+	static unsigned long stallmillis;
+	static long lastcoder;
     
     #ifdef FAKE_SERVO
       static unsigned long lastPrint = millis();
@@ -103,10 +105,29 @@ void   Axis::computePID(){
     if (_pidController.Compute()){
         // Only write output if the PID calculation was performed
         motorGearboxEncoder.write(_pidOutput);
+		
+		if (millis() - stallmillis > STALLTIMEOUT)
+		{
+			stallmillis = millis();
+			
+			long delta=lastcoder - motorGearboxEncoder.encoder.read(); 
+			
+			if(lastcoder == 0 || delta >  STALLSTEPS || delta < STALLSTEPS)
+			{
+				moved=lastcoder !=0;
+				lastcoder = motorGearboxEncoder.encoder.read();
+			}
+			else
+			{
+				stalled=1;
+				stalldelta= delta;
+			}
+		}
+    } else {
+      lastcoder=0;  //We've been deactivated; reset counts
     }
     
     motorGearboxEncoder.computePID();
-    
 }
 
 void   Axis::disablePositionPID(){
@@ -190,9 +211,8 @@ void   Axis::changeEncoderResolution(float *newResolution){
 }
 
 int    Axis::detach(){
-    
     motorGearboxEncoder.motor.detach();
-    
+	stalled=moved=0;
     return 1;
 }
 
@@ -300,3 +320,4 @@ void   Axis::test(){
 
 double  Axis::pidInput(){ return _pidInput * *_mmPerRotation;}
 double  Axis::pidOutput(){ return _pidOutput;}
+char	Axis::name(){ return _axisName;}
--- a/cnc_ctrl_v1/Axis.h
+++ b/cnc_ctrl_v1/Axis.h
@@ -53,8 +53,9 @@
             double     pidInput();
             double     pidOutput();
             long  steps();
-			char  moved, stalled;
-			long  stalldelta;
+			bool  moved;        //1 if the axis has ever moved since it's been attached
+            char  stalled;      //+1 every time through the PID loop if the encoder hasn't changed
+			long  stalldelta;   //Debug var keeping track of what the delta was (so we can fine tune STALLSTEPS if it shows up at wrong times.
 			char  name();
 			
         private:
--- a/cnc_ctrl_v1/Axis.h
+++ b/cnc_ctrl_v1/Axis.h
@@ -18,6 +18,9 @@
     #ifndef Axis_h
     #define Axis_h
 
+	#define STALLTIMEOUT	100 //100ms
+	#define STALLSTEPS		10	//Deadband; if it hasn't moved STALLSTEPS in STALLTIMEOUT, the axis is stalled.
+	
     class Axis{
         public:
             void   setup(const int& pwmPin, const int& directionPin1, const int& directionPin2, const int& encoderPin1, const int& encoderPin2, const char& axisName, const unsigned long& loopInterval);
@@ -50,7 +53,10 @@
             double     pidInput();
             double     pidOutput();
             long  steps();
-            
+			char  moved, stalled;
+			long  stalldelta;
+			char  name();
+			
         private:
             int        _PWMread(int pin);
             void       _writeFloat(const unsigned int& addr, const float& x);
@@ -66,5 +72,5 @@
             bool       _disableAxisForTesting = false;
             char       _axisName;
     };
-
+	
     #endif
--- a/cnc_ctrl_v1/Report.cpp
+++ b/cnc_ctrl_v1/Report.cpp
@@ -236,7 +236,20 @@ void  returnPoz(){
     static unsigned long lastRan = millis();
     
     if (millis() - lastRan > POSITIONTIMEOUT){
+        char tmpbuf[80], buflen=0;
+        memset(tmpbuf, 0, sizeof(tmpbuf));
         
+        for (int i=0;i<3;i++){
+            if(axes[i].stalled > 5){
+                buflen+=snprintf(tmpbuf+buflen, sizeof(tmpbuf)-buflen, "%c %s ", axes[i].name(), (axes[i].moved > 5) ? F("STALL") : F("UNHK "));
+            }
+        }
+        
+        if(buflen>0){
+            Serial.print(F("Message: "));
+            Serial.println(tmpbuf);
+            pause();    //Which should unattach the axes, which should clear the error.
+        }
         
         Serial.print(F("<Idle,MPos:"));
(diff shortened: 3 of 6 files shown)


davidelang:
what if an axis should not be moving (based on the speed/angle of the current 
move)?

thormj:
I was assuming that was being handled by the `if(_pidController.Compute())` statement, but that might doesn't look correct.
I put a `if(_pidOutput)` bit in there (so if 0 velocity is called for, reset things -- I was assuming that if 0 was called for, then Compute would be false, but that doesn't make a whole lotta sense - if that were true how would you stop an axis?)

krkeegan:
Couldn't this trigger a stall warning if the movements requested were extremely small?  When we are station keeping, the pidoutput may fluctuate rapidly positive and negative to try and keep holding the exact position.

cnc_ctrl_v1/Axis.h
```diff
@@ -50,7 +53,10 @@
             double     pidInput();
             double     pidOutput();
             long  steps();
-            
+			char  moved, stalled;
+			long  stalldelta;
```
krkeegan: Is this ever used?  I only see it being set a value, never being used as a value.

cnc_ctrl_v1/Report.cpp
```diff
@@ -236,7 +236,20 @@ void  returnPoz(){
     static unsigned long lastRan = millis();
     
     if (millis() - lastRan > POSITIONTIMEOUT){
+        char tmpbuf[80], buflen=0;
+        memset(tmpbuf, 0, sizeof(tmpbuf));
         
+        for (int i=0;i<3;i++){
+            if(axes[i].stalled > 5){
+                buflen+=snprintf(tmpbuf+buflen, sizeof(tmpbuf)-buflen, "%c %s ", axes[i].name(), (axes[i].moved > 5) ? F("STALL") : F("UNHK "));
```
krkeegan: This is nice and compact but hard to read.  I would be inclined to waste program space, which we will have a good amount of, to make this easier for humans to comprehend.

Also, it doesn't seem to me that moved has the value that you are testing for here.  I think you want moved to have the value of delta maybe?

cnc_ctrl_v1/Report.cpp
```diff
@@ -236,7 +236,20 @@ void  returnPoz(){
     static unsigned long lastRan = millis();
     
     if (millis() - lastRan > POSITIONTIMEOUT){
+        char tmpbuf[80], buflen=0;
+        memset(tmpbuf, 0, sizeof(tmpbuf));
         
+        for (int i=0;i<3;i++){
+            if(axes[i].stalled > 5){
```
krkeegan: why greater than 5?  It doesn't look like stalled can ever be anything by 0 or 1?

cnc_ctrl_v1/Axis.cpp
```diff
@@ -103,10 +105,32 @@ void   Axis::computePID(){
...
+        if(_pidOutput) {
+            if (millis() - stallmillis > STALLTIMEOUT)
+            {
+                stallmillis = millis();
+                
+                long delta=lastcoder - motorGearboxEncoder.encoder.read(); 
+                
+                if(lastcoder == 0 || delta >  STALLSTEPS || delta < STALLSTEPS)
+                {
+                    moved=lastcoder !=0;
```
krkeegan: I don't understand this.  Is this a form of a ternary I wasn't aware of does it read:

set moved to equal lastcoder if lastcoder is not 0?

If so, this is a problem.  moved is a char and lastcoder is a long.  This will almost certainly overflow.

cnc_ctrl_v1/Axis.cpp
```diff
@@ -103,10 +105,32 @@ void   Axis::computePID(){
...
+                    lastcoder = motorGearboxEncoder.encoder.read();
```
krkeegan: Lastcoder needs to be an unsigned long if you are going to do this.

cnc_ctrl_v1/Axis.h
```diff
@@ -18,6 +18,9 @@
     #ifndef Axis_h
     #define Axis_h
 
+	#define STALLTIMEOUT	100 //100ms
+	#define STALLSTEPS		10	//Deadband; if it hasn't moved STALLSTEPS in STALLTIMEOUT, the axis is stalled.
```
krkeegan: :+1:

cnc_ctrl_v1/Report.cpp
```diff
@@ -236,7 +236,20 @@ void  returnPoz(){
     static unsigned long lastRan = millis();
     
     if (millis() - lastRan > POSITIONTIMEOUT){
+        char tmpbuf[80], buflen=0;
+        memset(tmpbuf, 0, sizeof(tmpbuf));
```
krkeegan: I don't know if this is ideal, creating and deleting a 160 byte array 5 times a second.

cnc_ctrl_v1/Report.cpp
```diff
@@ -236,7 +236,20 @@ void  returnPoz(){
...
         
+        for (int i=0;i<3;i++){
+            if(axes[i].stalled > 5){
+                buflen+=snprintf(tmpbuf+buflen, sizeof(tmpbuf)-buflen, "%c %s ", axes[i].name(), (axes[i].moved > 5) ? F("STALL") : F("UNHK "));
+            }
+        }
+        
+        if(buflen>0){
+            Serial.print(F("Message: "));
```
krkeegan: My preference would be to use the reportAlarmMessage or reportFeedbackMessage functions instead of this.  I kinda feel like this is an alarm instance.  

I think these functions will give us more flexibility in the future (other languages, only reporting error numbers to save program space).

cnc_ctrl_v1/System.h
```diff
@@ -71,6 +71,7 @@ extern system_t sys;
 extern Axis leftAxis;
 extern Axis rightAxis;
 extern Axis zAxis;
+extern Axis axes[3];
```
krkeegan: I don't know enough about c++, how is something like this stored?  How much memory does this array consume?  Is it only a reference to the existing axes or does it contain a duplicate instance?

thormj:
I tried to mitigate that with thresholds ("you have to have indicated a stall for the last X counts of the `report` all"), but that still possible... is there a "distance count" (preferable in encoder counts --  that should be used to indicate "station keeping" vs "move me" -- I'll add it /change it in  `Axis.cpp:116` and then we'd be sure (or `Axis.h` -- I chose StallSteps and StallTimeout quite arbitrarily)

cnc_ctrl_v1/Axis.cpp
```diff
@@ -103,10 +105,32 @@ void   Axis::computePID(){
...
+                }
+                else
```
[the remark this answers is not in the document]
thormj: That was my _assumption_ - that if stalled, the encoder would not change at all.  Let me know if we need to put a bit of "if it moves by only 2 counts, it's still stalled."

cnc_ctrl_v1/Axis.cpp
```diff
@@ -83,6 +83,8 @@ void   Axis::setSteps(const long& steps){
 }
 
 void   Axis::computePID(){
+	static unsigned long stallmillis;
+	static long lastcoder;
```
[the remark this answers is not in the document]
thormj: Good catch.

cnc_ctrl_v1/Report.cpp
thormj: Yeah, I goofed.  Moved is a boolean.  If it's ever moved, it's stalled.  Otherwise, it's unhooked.

cnc_ctrl_v1/Report.cpp
thormj: Err... oops.  That's why I figured it should be a char (0-240 if report doesn't get called often. Fixed.

cnc_ctrl_v1/Axis.cpp
thormj: No, it's set moved to `if lastcoder is not zero` (so if lastcoder=0 -- the initial run of the function (otherwise it'll be the last encoder value), then moved=0.  OTOH, if delta is outside of STALLSTEPS, then it /has moved/.

davidelang:
I'm not sure that looking for no motion is the right thing to be doing, the 
current problem of a hook at the end of a line at the top of the machine is a 
classic example of the machine not keeping up (what we would consider the 
equivalent of a stall), but motion was happening, just not enough of it

So I think we need to be looking for too much error, not no motion

cnc_ctrl_v1/Report.cpp
thormj: Now... that seems silly.  It's one of those things where I thought I was being nice and encapsulated, but it does seem like a lot of bull... fixed.

davidelang:
or another thing to look for, maxing out the motor power and the PID loop still trying to increase it.

cnc_ctrl_v1/Report.cpp
thormj: Sounds like a good idea.  I borrowed half of that idea mainly because I want to see what the debug value is in case of a stall...

cnc_ctrl_v1/System.h
thormj: Good point... unfortunately I don't know either, so... I forced it to become an array of pointers to the real axes...

krkeegan:
I think we had talked about using error before and that seems a bit more straight forward to me.  The only issue I see is that if the max_error is a fixed amount, small movements would not trigger the warning, at least not immediately.  The max_error could scale but then it is a question of what does it scale with?

I think this is a case where a simple solution fixes many of our problems.  But we can't fix all of the issues without a complex solution.

cnc_ctrl_v1/Axis.cpp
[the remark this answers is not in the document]
krkeegan: ahh, the negative value clears things up.  Ok that makes sense to me now.  basically a check of an abs value.

cnc_ctrl_v1/System.h
```diff
@@ -71,7 +71,7 @@ extern system_t sys;
 extern Axis leftAxis;
 extern Axis rightAxis;
 extern Axis zAxis;
-extern Axis axes[3];
+extern Axis *axes[3];
```
krkeegan: :+1:

krkeegan:
@thormj thank you for making those corrections.

If this works as advertised and produces no side effects I think this is good to commit.  I can keep backseat driving about hypothetical alternative ways to handle this, but I lack the time to actually do the coding, so that doesn't seem fair.

thormj:
I might've been trying to shoot bigger game than I should've been.  The thing that _I wanted_ was a way to burp and tell me "hey bozo, plug in your Z axis!", but I figured that that would've been the same algorithm for detecting a stall (I moved before, but for some reason, I cannot move now), so I said "if I've moved before... I'm plugged in; say that I've stalled instead."

It doesn't detect "I'm not keeping up" very well, and I'm not sure if will actually detect a stall or not (if during a stall, the encoder bounces more than 10 clicks per 100 ms, it _not-quite stalled_ by my code's reckoning.

Maybe for "not keeping up" you'd want an error and all stop ("Hey idgit - plug in yo' z axis"), but I think you'd really rather  start tweaking the other velocities in a coordinated move to straighten it back out and set a flag that "your Maslow is cutting slow because it's binding" (not good for cutters [but would signalling a full stop without backing off the material be good either?], but  at least the cut will be good and straight).

Maybe if vel_error>2000, cut velocities of all axes to 1/2, and if it gets to within 100, go back to full speed?  But I wouldn't try to make that kind of change away from maslow as I have no idea what the "usual" and the "OMG-it's sticking" numbers would be.

davidelang:
On Wed, 21 Feb 2018, Kevin Robert Keegan wrote:

> I think we had talked about using error before and that seems a bit more 
> straight forward to me.  The only issue I see is that if the max_error is a 
> fixed amount, small movements would not trigger the warning, at least not 
> immediately.  The max_error could scale but then it is a question of what does 
> it scale with?


> I think this is a case where a simple solution fixes many of our problems. 
> But we can't fix all of the issues without a complex solution.

I don't think most of our problems are with no movement.

That's why I was suggesting looking for exceeding +-255 speed. Even if it's a 
small movement, if we aren't moving, the PID loop should keep calling for more 
speed up past the limit, because the error isn't being reduced.

blurfl:
Wouldn’t overshoot on a poorly tuned PID loop cause this error?

davidelang:
only if the overshoot is bad enough to cause 100% power in reverse to be 
applied, and if that does happen, I don't think it's a bad thing to get a notice 
of it.

The notice should be something like "the machine was unable to move at the 
desired speed" (or something along those lines), not "motor stalled" or 
something like that.

David Lang

blurfl:
How much overshoot is reasonable? For how long?

davidelang:
ideally we should not overshoot (it means cutting areas we aren't supposed to 
cut)

currently some overshoot is going to happen because we do not have any motion 
planning to slow us down. I don't know what this amounts to, but it's not enough 
to notice in the resulting cuts.

k

krkeegan:
We have to be careful about what we mean when we say overshoot.

Overshoot in reference to the PID loops means requesting either an RPM or voltage value higher than necessary.  The issue is we don't know what is necessary. 

@blurfl is correct, a badly tuned PID loop would cause this type of design to fire.  But I agree with @davidelang that above some threshold, this is probably what we want anyways.  

I think the only way to pick that threshold is by some empirical testing.  Of course the problem there is that a different design may result in a different outcome.

This proposal is certainly simpler and cleaner than the one currently designed

davidelang:
I'm proposing that the first-try threshold be the max power to the motors. If we 
are running the motors at full power and we think (through the PID loop) that we 
need to go even faster, that seems like a clear problem

either:
1. we aren't moving
2. we aren't keeping up with the desired movement
3. we have passed (by a significant amount)where we were supposed to stop and 
are reversing in an attempt to get back where we are supposed to be.

I don't think 3 is common (but real-world experimentation will tell us)

I am hoping that this test will trigger for the heavy-sled-top-of-machine 
problem that gero ran into.

David Lang

krkeegan:
It will be interesting to see what happens.  My guess is that when starting from a standstill, that we hit your 2nd option (not keeping up) and max out often.  Here is a graph I made a while back

https://discourse-cdn-sjc2.com/standard11/uploads/maslowcnc/original/2X/7/7c5e44c3703fa92a1d624342b8fd5b4ee35c8cdc.png

It looks like it may not hit 255, but that could just be because the sample rate of the graph is well below the PID loop frequency.

davidelang:
yep, let's see what happens.

If this isn't hit under normal operations, this would be a very useful check to 
have.

if it is, then we need to implement acceleration planning before it is as 
useful.
k

krkeegan:
So I think @davidelang is right.  I made a simple test branch that prints a warning to the serial stream if 

255 < speed < -255

and this condition is maintained for 20 milliseconds (meaning 3 consecutive PId cycles).

You can try it here:

https://github.com/krkeegan/Firmware/tree/OverSpeedWarning

It works.  It shows up when i unplug a motor.  And it also prints on the areas on the work surface where my motors are not able to maintain the 1000mm/min max speed.  (top corners mostly, I don't know if this is unique to my machine or if we have our max speed set a little high).

In this basic form it doesn't understand a stall vs unhooked, but I don't know if that is necessary.

Anyways, others should play around with this and see if this works for them.  We can implement it as a warning very easily, if we want to actually trip an alarm state that still needs more work to enable alarms to work smoothly.

blurfl:
@krkeegan, are you considering a PR with the overspeed warning in the simple form you linked? I could see making it switched by verboseDebug, and maybe making the trigger something like (LOOPINTERVAL * 2)/1000 to scale with changes to the PID loop.

davidelang:
the current check to alert if at max for more than 3 cycles is a good starting 
point.

I think 'alert if at max and PID loop thinks we need to speed up' will work as 
well, and doesn't have the complexity of remembering the last few cycles (or 
worrying about how far apart those cycles are)

but either one is going to give false positives until we get acceleration 
planning in place.

davidelang:
also check out this Marlin issue 
https://forums.maslowcnc.com/t/calibration-moving-one-motor-wrong-direction-in-ground-control-v1-08/2590

BarbourSmith:
I would like to get this into 1.09. I'm a little hesitant because we accidentally put pretty significant bugs in 1.06 (crashed when velocity couldn't be reported) and 1.07 (crashed when setting the chain lengths) so we haven't had a really stable release in like a month, but it sounds like there is a strong community desire to see this feature added.

How do I test that it is working? When I unplug a motor I don't see any sort of warning.

blurfl:
I think that the change @krkeegan wrote [here](https://github.com/krkeegan/Firmware/tree/OverSpeedWarning) is simpler and easier to test. It posts a warning message if the motor becomes disconnected or when the PID loop asks for more than 100% effort.

BarbourSmith:
Excellent! 

@krkeegan how do you feel about making a PR with those changes?

krkeegan:
I can do that, this is very simple and is probably a good baby step, since it only produces a warning for the user.

Once we get it dialed in, and other improvements, it could actually take action (ie stop) if a fault happened.

BarbourSmith:
I'm going to close this one as added in #410 👍 👍

Issue #363, opened 2018-01-06
Create AVR Simulated Project for Virtualized Testing of Firmware


krkeegan:
SimAVR is an AVR simulator that can be used for testing of the firmware.  

https://github.com/buserror/simavr

I am sure other avr simulators are also available.

krkeegan:
I don't think so.  Processing looks like a robust means to communicate with an Arduino, but I don't think it would allow the simulation of the firmware on a PC.

idoc:
Simavr seems very promising, but has anyone gotten it to run any Arduino code successfully? I've spent about 15 hours so far trying to get anything to run for more than a few seconds, and it works, sort of, until I try to write to the serial port. Once I wrote more than a few hundred bytes to serial, the whole thing hangs. This happens even with trivial test programs. Maslow's firmware is chatty enough that it hangs before I even get a chance to open the serial monitor :)

I've opened an issue on the simavr repo. Hopefully it's just some boneheaded mistake  I'm making, and someone there can point it out.

idoc:
I have…

Something.

I (finally!) got most of the firmware running in simavr. It required much futzing with simavr code, but most of our code runs now, and I have  a simple recipe that others can copy-paste into a Mac or Linux terminal.

The issue is that it only runs if I disable the timer that runs runsOnATimer. Something about how we use interrupts in TimerOne.cpp just makes simavr hang after a few seconds.

I still think simavr has a decent chance of giving us a solution for trying out code without a spare Arduino, having unit tests that run automatically in a semi-realistic environment (a step towards a full CI environment?), and being able to attach gdb to the firmware and seeing what's going on inside.

I'm looking at two directions now:	

1. See if I can find an alternative timer implementation, one which works better in simavr. If I find one that works, I'll add an option in `Config.h` to use it when testing.

1. See if I can completely avoid using a timer in testing. Maybe call `runsOnATimer()` directly inside `loop()`? Maybe generate fake PID values whenever someone tries to read them?

Option 2 seems like it will require less reverse engineering of both simavr and Maslow, and will allow me to have something useful to share with everyone sooner.

To be honest, I only have a general idea of what `computePID()` is for. I'm guessing that in a simulator, it won't be a problem if it does run often enough, or if it returns inaccurate values, because there's no real sled or real motors to keep accurate track of.

Does this seem like a valid assumption that will give us useful simulation? Any advice would be much appreciated.

BarbourSmith:
[...]

PR #395, opened 2018-02-19
Add a PlatformIO project

### LINKED ISSUE #365: Feature Request: makefile build process of firmware (for non-x KivyPi setup)
I'm enamored how well GroundControl with KivyPi runs without Xorg on the Raspberry Pi.  Then I got to thinking, "I'm still going to need to use Xorg for compiling and uploading the Arduino's firmware with the IDE".  After reading the CLI section on arduino.cc, they present a simple makefile process to build and upload firmwares without the full blown IDE.  How cool would it be to include a makefile configuration that pulls the latest code from the git repo, build and uploads?  No need for Xorg :-)

As soon as the Maslow kit arrives in the mail, I'll get started on putting this makefile config together, unless someone beats me to it :-)

idoc:
This project builds for two environments:
megaatmega2560: This is the standard Arduino Mega 2560 setup for
PlatformIO. It uses the default flags for PlatformIO, and builds an
optimized, stripped binary. Good for building firmware for production
use.
simavr: Disable all optimizations, leave debugging symbols, and #define
SIMAVR. SIMAVR doesn't do anything yet, but it will enable some
simavr-specific behavior in a separate commit.

I created this project as prep for #363, but it can also help simplify
setup for users of the PlatformIO IDE, or someone setting up a CI
environment.

I think it also resolves #365.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,13 @@
 .clang_complete
 .gcc-flags.json
 .travis.yml
-platformio.ini
 lib/readme.txt
 cnc_ctrl_v1/.DS_Store
 cnc_ctrl_v1/.DS_Store
 cnc_ctrl_v1/.DS_Store
 .DS_Store
 **/.DS_Store
+.vscode/
+.vscode/c_cpp_properties.json
+.vscode/launch.json
+simduino_atmega2560_flash.bin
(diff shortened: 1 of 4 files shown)


BarbourSmith:
Excellent! I am sure it will help someone with their setup soon. 👍 👍

blurfl:
Where would be a good place to make the info in the initial post of this issue available for posterity?

PR #396, opened 2018-02-19
Add support for running in simavr

idoc:
Simavr is a simulator for the AVR family of microcontrollers
(https://github.com/buserror/simavr). Running Maslow firmware in simavr
is useful for automated testing in a CI loop and debugging in GDB.

This is the next step in #363.

These commit works around two issues with simavr:
1. The simulated UART crashes if you don't flush often, so we flush
after every character written to Serial.
2. Our TimerOne class causes the simulator to hang, so I moved the call
to runsOnATimer() into loop(). This means runsOnATimer() will be called
less frequently, but I think that's OK for the simulated environment.

### DIFF
--- a/cnc_ctrl_v1/Config.h
+++ b/cnc_ctrl_v1/Config.h
@@ -29,6 +29,9 @@
                            // a servo updating the encoder steps even if no servo
                            // is connected.  Useful for testing on an arduino only
 
+// #define SIMAVR          // Uncomment this if you plan to run the Firmware in the simavr
+                           // simulator. Normally, you would not define this directly, but
+                           // use PlatformIO to build the simavr environment.
 
 #define LOOPINTERVAL 10000 // What is the frequency of the PID loop in microseconds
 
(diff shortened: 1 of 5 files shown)


BarbourSmith:
Looks great! Its fun to see how quickly this is progressing 👍 👍

blurfl:
Under normal non-simavr use does the [env:simavr] section in platformio.ini need to be commented out to prevent platformio from building and uploading to the mega? 
>Looking for upload port...
Auto-detected: /dev/cu.usbmodem1411
Uploading .pioenvs/simavr/firmware.hex

idoc:
Oh dang. I was so focused on getting simavr to work, I totally forgot to
test the full workflow on a real Arduino...

I have a PR to disable the upload for simavr. I still need to test it with
an Arduino plugged in, so I won’t be able to upload it till I get home this
evening.

blurfl:
Thanks for the PlatformIO setup, it offers some interesting options. I'm also interested in getting a simavr environment set up for the Teensy 3.5 & 3.6 boards. Getting PlatformIO set up for those was easy, and commenting out the unwanted sections of platformio.ini seemed to be the way to 'steer' the operation.

blurfl:
I'll test the PR and report. I'd be glad to use a script, it sounds like a good idea.

PR #403, opened 2018-02-22
Disable builds for Teensy boards

idoc:
I accidentally mixed a change that was meant for everyone, with a change
I created for @blurfl to test, and now if anyone tries to use the
PlatformIO project, it will try to build a Teensy version of the code,
which won't work with the current tip of master.


idoc:
Sorry about that...

blurfl:
I'm sorry I missed it as well.

PR #408, opened 2018-02-27
Setup Axes Again After Changing Chain Rotation Direction

krkeegan:
Means we don't have to restart.


krkeegan:
Should be merged with

https://github.com/MaslowCNC/GroundControl/pull/658

No version number change required.

BarbourSmith:
Looking good! I'm really glad we got this in today because that means 1.07 will include the no need to restart! 👍 👍

PR #410, opened 2018-03-07
Over Speed Warning

krkeegan:
Outputs a simple warning if the motors appear to be disconnected or unable to keep up.


cnc_ctrl_v1/Report.cpp
```diff
@@ -71,6 +71,8 @@ void  reportStatusMessage(byte status_code){
           // Serial.println(F("Unsupported command")); break;
           // case STATUS_GCODE_UNDEFINED_FEED_RATE:
           // Serial.println(F("Undefined feed rate")); break;
+          case STATUS_OVERSPEED:
+            Serial.println(F("OVERSPEED WARNING: Check feedrate and motors.")); break;
```
BarbourSmith: I like this. I'm glad it is its own warning instead of going on the "Message: " command which would clutter the screen with popups.

PR #414, opened 2018-03-22
gcc warnings Part III

aprospero:
The (hopefully) final pull request regarding this issue.

Platformio build does not benefit from activating compiler warnings, since it triggers a lot of them in board libray code by the disabled compiler optimization. So, switching them on is reverted.

Another unused Variable was found and deleted.

That's it, folks! :)


aprospero:
Haha,

   I just realized that platform.io already compiles with warnings on. So it was right to revert the commit, but I did it out of false assumptions.  Luck is with the oblivious :dancer:

Issue #417, opened 2018-03-26
AUX pins should be defined at runtime


blurfl:
Various versions of PowerControl board have had different AUX pin connections. Presently they are configured using #define in Config.h  To date we have used the AUX pins which did not change between versions of hardware.
 In order to make use of all the AUX pins  including the ones that change between boards, it seems like they should be defined at runtime based on the board version much as the INx, ENx and ENCODERxy pins are.
 Is there a clean way to do this without needing to re-write Spindle and Probe as classes in the way that Axis is done?

BarbourSmith:
Hmmm that's a good question.

I agree that the AUX pins should be defined at run time because we will want to move them in the future.

From a quick look at the code I can't see a way to do it other than how the axies are done, but it looks like both of those classes were designed with the pins as variables so it shouldn't be too hard to switch them over

blurfl:
I see that _I_ was the one who originally #defined the AUX pins in PR #223 almost a year ago (blush, shame) and it didn't account for AUX5 and AUX6.
 I agree that the class approach is doable, I just hoped for a simpler one.

aprospero:
To be honest, the Probe files would have been on my list for deleting to simplify the program flow (see #415). The single only function just wraps another function call...

blurfl:
Not stupid, but your effort could take us further toward that goal as well. Refactoring our code to make grbl sections 'drop in' could be a big advantage.

aprospero:
Agreed. And this could be the solution for finding common Coding guidelines as well. We could take grbl as blue print. their code seems to be well formatted - at least on first glance.

aprospero:
Hmmm,
maybe I should ask this elsewhere, in the forum or in the wiki. How far are you planning to adopt the grbl code structure?
In my understanding the math of maslowCNC is completely different to grbl, starting with servo vs. stepper and not ending with the peculiarities of the maslow sled. A lot of layer concepts like the planner or the protocol layer could be rewritten though, but I doubt that we can just copy paste code in a big way.
If this is correct and we have to rewrite most of the code anyway, why should we copy the grbl code structure? For sure, they did their homework - but that doesn't mean that their concepts are in every detail as good for us as for them.
I can imagine that the idea was appealing when it came to adapting gcode parsing, but in my understanding that task is mostly done and maslow has already most important/necessary gcodes adapted.
So what is the benefit of using grbl as blueprint?

Just want to understand the intention...

blurfl:
[...]

PR #419, opened 2018-03-28
Define aux pins at runtime

blurfl:
This addresses issue #417 .
In order to make use of all the AUX pins including the ones that change between board versions, it seems like they should be defined at runtime based on the board version much as the INx, ENx and ENCODERxy pins are. This code does that, moving them into System.cpp to be defined with those others. I've also added comments to Probe.cpp and Spindle.cpp to tell where the active pins are assigned.

Note 1, I've left the debug print statements in until testers have evaluated the PR. The 'dummyPin' uses an AUX pin that differs between board versions. The v1.0 board should show '1', the v1.1 and v1.2 boards should show '60'.

Note 2, I've created a function to assign the AUX pins that are currently used to the variables that need them. The function takes all the AUX pins values as parameters. This was done to allow defining them without raising the compiler complaint about unused #defines. I feel having the AUX pins defined even if they are not currently used will make it easier for others to make use of them in the future.


BarbourSmith:
Wonderful!!!

I will test this against all of the PCB versions tomorrow and report back what it reports back 😀

BarbourSmith:
I tested this against version 1.0, 1.1, and 1.2b of the hardware and they all look good. 

I did not test every aux pin for every version because I didn't really have a good way to do that, but we can be sure that this will at the very least not break any existing functionality.

I have full compiler warnings turned on, and when compiling the firmware I did see a new warning:

```
Archiving built core (caching) in: C:\Users\Bar\AppData\Local\Temp\arduino_cache_196560\core\core_arduino_avr_mega_cpu_atmega2560_a94ab6aaf61dfb93b4a8079c694a14c2.a
sketch\System.cpp: In function 'setupAxes':

sketch\System.cpp:214:18: warning: 'AUX6' may be used uninitialized in this function [-Wmaybe-uninitialized]

   dummyPin = AUX6;
```

I think we're ready to merge. @blurfl do you want to take out some of the print statements before I do?

blurfl:
Yes, I'll take the print statements and the dummy pin out. Could it wait until tomorrow?

BarbourSmith:
Absolutely! No rush

BarbourSmith:
Looks good to merge!! 👍 👍

PR #422, opened 2018-04-06
Add chain tolerance parameters

reecej:
This adds chain tolerance parameters for the left and right chains. These parameters allow the chain pitch to be fine tuned for each individual chain.

This is a first step to allowing chain tolerance calibration. In a future pull request the calibration procedure will be updated to include chain tolerance for both chains. Leaving these new values set to zero should not impact existing users. Changing the value will require a chain length calibration.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -213,23 +213,23 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
 
     //Calculate the chain angles from horizontal, based on if the chain connects to the sled from the top or bottom of the sprocket
     if(sysSettings.chainOverSprocket == 1){
-        Chain1Angle = asin((_yCordOfMotor - yTarget)/Motor1Distance) + asin(R/Motor1Distance);
-        Chain2Angle = asin((_yCordOfMotor - yTarget)/Motor2Distance) + asin(R/Motor2Distance);
+        Chain1Angle = asin((_yCordOfMotor - yTarget)/Motor1Distance) + asin(RleftChainTolerance/Motor1Distance);
+        Chain2Angle = asin((_yCordOfMotor - yTarget)/Motor2Distance) + asin(RrightChainTolerance/Motor2Distance);
 
-        Chain1AroundSprocket = R * Chain1Angle;
-        Chain2AroundSprocket = R * Chain2Angle;
+        Chain1AroundSprocket = RleftChainTolerance * Chain1Angle;
+        Chain2AroundSprocket = RrightChainTolerance * Chain2Angle;
     }
     else{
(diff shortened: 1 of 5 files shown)


BarbourSmith:
This is a fantastic idea! Excellent work!! 👍 👍 

Just out of curiosity is there a reason we couldn't just use the chain tolerance variable to tweak the chain pitch variable?

reecej:
There are a couple aspects to this. For one, the chain pitch itself isn't sent to the firmware. Rather, it's combined with the sprocket tooth count, and then the total distPerRotation is sent. We could adjust this total distPerRotation, but that would not allow us to calibrate the left and right chains individually.

Also, I kept the original, unmodified distPerRotation as it is used elsewhere, such as the quadrilateral kinematics.

davidelang:
On Fri, 6 Apr 2018, BarbourSmith wrote:

> Just out of curiosity is there a reason we couldn't just use the chain tolerance variable to tweak the chain pitch variable?

It may not be the same for both chains, and we may have some place (like around 
the sprockets) where the real chain pitch applies, and then other places where 
the 'stretched' version applies.

BarbourSmith:
Great points @reecej 👍 👍

BarbourSmith:
@reecej Some folks have been reporting strange behavior when setting these where they get the sprockets to reverse direction and behave in unexpected ways. I'm trying to dig into it but I'm not sure I fully understand what is going on.

Can you help me understand how this works?

It looks like we're recomputing the radius based on the chain tolerance value?

BarbourSmith:
It looks to me like there might be an issue in the way the value for 

`kinematics.RrightChainTolerance = (sysSettings.distPerRotRightChainTolerance)/(2.0 * 3.14159);`

is computed from the dist per rot sent up.

Here's what I'm seeing when I add some debug code:

image

If I set the chain tolerance to .2 in GC it ends up as 10.13 in the firmware...does that seem right? Should there be a conversion between those two?

reecej:
The values you entered seem to match the results displayed there. The chain tolerance is meant to be a percentage added length over the nominal, so entering 0.2 means each link would be 6.35 * 1.002 = 6.3627 mm in length. This would equate to a 63.627 mm circumference for a 10-tooth sprocket, giving an effective radius of 10.127 mm.

I'll have to dig into this more. My apologies for the absence, I have had some other obligations come up that have greatly limited my free time. I'll review and see what I can find.

blurfl:
I'll be interested to hear what you find🙂. I wonder if some other problems we've had are related to the values returned by readFloat.

BarbourSmith:
That's a good thought!

I wonder if it is giving us other strange behavior. I know that the issue I miss diagnosed as an issue with the motor keeping up detection code this morning was caused by this issue so it's certainly possible that it's giving us trouble elsewhere

PR #423, opened 2018-04-07
3D, and/or 3-Axis movements

mrfugu:
Bumping this up on the list for the following reasoning: https://forums.maslowcnc.com/t/how-fast-does-everyone-cut/3172/6

based on the age of this code, I'm sure it'll need a complete go through/review from someone more competent than I, but I think that ramping in and out of cuts, maintaining a constant load on the router, sled and workpiece will result in better loading on the router, better cuts and faster processes.

### DIFF
--- a/cnc_ctrl_v1/CNC_Functions.h
+++ b/cnc_ctrl_v1/CNC_Functions.h
@@ -399,7 +399,7 @@ float calculateFeedrate(const float& stepSizeMM, const float& msPerStep){
     
     float feedrate = (stepSizeMM*MINUTEINMS)/msPerStep;
     
-    return feedrate;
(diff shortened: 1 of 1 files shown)


blurfl:
Shouldn't this be an issue instead? Much has changed since this code was mergable, it really is no longer a PR.

mrfugu:
yeah, most likely, I really just wanted to look for any previous mention of 3-axis movement and found this branch.
I figured starting here since *some* work has been completed already. 

Either way, true 3-axis movement, should improve accuracy by allowing more control over bit loading, at least with the CAM applications that are able to take advantage of that. (ie: Fusion360)

PR #425, opened 2018-04-08
Add G4 gcode, use maslowDelay() to remain responsive to GC during dwell time

blurfl:
Dwell pauses the command queue and waits for a period of time. Prints 'dwell time XX' to the display. The Hold and Stop buttons remain available during the dwell time.

Usage:
G4 [P<time in ms>] [S<time in sec>] (only one or the other is allowed, Sxx is ignored if Pxx is present)


cnc_ctrl_v1/GCode.cpp
```diff
@@ -706,22 +706,25 @@ void  G4(const String& readString){
       The G4() dwell function handles the G4 gcode which pauses for P milliseconds or S seconds.
       Only one of the two is accepted, the other ignored.
       Use maslowDelay() to remain responsive to GC during the dwell time.
+      Because maslowDelay() operates in milliseconds, round to the nearest millisecond.
+      Negative values are treated as positive (not a time machine).
```
BarbourSmith: Do we want to open an issue for a time machine feature? 😁

BarbourSmith:
Great work! 👍 👍 🎉

cnc_ctrl_v1/GCode.cpp
blurfl: >Do we want to open an issue for a time machine feature?

only if we can make it go forward, to where it's all working just right 😁

PR #427, opened 2018-04-13
remove function that was replaced in PR #389

blurfl:
PR #389 replaced this function
>void setPWMPrescalers)

 with one of the same name 
>void setPWMPrescalers(int prescalerChoice)

which accepts a setting parameter. This PR removes the unused version of the function.


BarbourSmith:
Excellent keeping things clean! 👍 👍 🎉

Issue #426, opened 2018-04-09
This file does not cut correctly


BarbourSmith:
When cut this file does not cut correctly

[test cut square rounded corner.txt](https://github.com/MaslowCNC/Firmware/files/1891189/test.cut.square.rounded.corner.txt)

You can see that there is a bump over on the left hand side:

![maslow miscut 2](https://user-images.githubusercontent.com/9359447/38514744-e1d2fb2a-3be6-11e8-9456-344b708c3847.jpg)

While it renders right in Ground Control:

![maslow miscut](https://user-images.githubusercontent.com/9359447/38514773-f8192e18-3be6-11e8-823d-8caf04f4556c.jpg)

Here is the log for a test cut of the file:

[log.txt](https://github.com/MaslowCNC/Firmware/files/1891194/log.txt)

And the forums conversation about the issue can be found here:

https://forums.maslowcnc.com/t/odd-cut-when-cutting-test-shape/3230/3

BarbourSmith:
That sounds like an overflow issue in the arduino to me, or a rounding issue when dealing with numbers that large

davidelang:
well, it's that and a bad design file (it should not be doing arcs instead of 
stright lines)

davidelang:
On Mon, 9 Apr 2018, Scott Smith wrote:

>> That sounds like an overflow issue in the arduino to me, or a rounding issue when dealing with numbers that large
>
> Looking at the gcode parameter values as they are handled in G2(), I see
> float extractGcodeValue(const String& readString, char target, const float& defaultReturn)
> returning more digits than were in the source gcode string. I added code to G2() to print the I and J values returned by extractGcodeValue() as rawI and rawJ:
>
> Sent: G3 X-15.1990 Y-10.880 I9563.8560000000 J0
> iCS executing G code: G3 X-15.1990 Y-10.880 I9563.8560000000 J0
> rawI 9563.855468750000
> rawJ 0.000000000000
>
> Haven't yet looked at why that might be, or whether it is related to this issue though.

That is floating point to fixed point precision loss. floating point doesn't 
keep that many bits of precision.

blurfl:
>That is floating point to fixed point precision loss

 float extractGcodeValue() should return a float, yes? Is it the case that the print statement doesn't correctly print it? Subsequent calculations seem to indicate that the extra digits are there.

BarbourSmith:
[...]

PR #428, opened 2018-04-13
Add an alarm for gcode calculation error

blurfl:
Add flag ALARM_GCODE_PARAM_ERROR and the code to handle it in Report.cpp
Use that flag in Motion.cpp if a parameter value error causes a calculation error when calculating an arc() as used by G2() in GCode.cpp
 Addresses Issue #426, and requires GC PR #697 for the alarm to trigger a dialog. The dialog:

alarm dialog

This .nc file will trigger the calculation error when the firmware evaluates line 31. 
[G2-G3-error-arises-in-line-31.nc.zip](https://github.com/MaslowCNC/Firmware/files/1908700/G2-G3-error-arises-in-line-31.nc.zip)

### DIFF
--- a/cnc_ctrl_v1/Motion.cpp
+++ b/cnc_ctrl_v1/Motion.cpp
@@ -225,6 +225,9 @@ void  singleAxisMove(Axis* axis, const float& endPos, const float& MMPerMin){
     
 }
 
+// return the sign of the parameter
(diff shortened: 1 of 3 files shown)


BarbourSmith:
Excellent!!!

PR #429, opened 2018-04-16
Add a system setting for positionErrorLimit, stop cutting if reached

blurfl:
After each time that position error is reported, check against a limit and take action if the limit is reached.

 - add $42 positionErrorLimit with a default of 2mm
 - check each time [PE: is sent
 - set sys.stop flag if left- or right-axis error >= positionErrorLimit

Note, this leaves GC with a 'Hold' in place rather than a 'Stop'. This leaves the line number visible.The user is asked to click the 'Stop' button to get GC into the same state as the firmware.

GC PR #703 makes an Advanced Setting available to change the value of positionErrorLimit.

### DIFF
--- a/cnc_ctrl_v1/Report.cpp
+++ b/cnc_ctrl_v1/Report.cpp
@@ -125,6 +125,11 @@ void  reportAlarmMessage(byte alarm_code) {
         pause(); //Now wait for user decision about 'Stop' or 'Resume'
         break;
         }
+      case ALARM_POSITION_LIMIT_ERROR: {
+        Serial.println(F("The sled is not keeping up with its expected position - make a note of the line number. Then click the 'Stop' button to clear the alarm.  "));
+        sys.stop = true;
+        break;
+        }
     }
   #endif
 }
@@ -176,6 +181,7 @@ void reportMaslowSettings() {
     Serial.print(F("$39=")); Serial.println(sysSettings.fPWM);
     Serial.print(F("$40=")); Serial.println(sysSettings.distPerRotLeftChainTolerance, 8);
     Serial.print(F("$41=")); Serial.println(sysSettings.distPerRotRightChainTolerance, 8);
+    Serial.print(F("$42=")); Serial.println(sysSettings.positionErrorLimit, 8);
     
   #else
     Serial.print(F("$0=")); Serial.print(sysSettings.machineWidth);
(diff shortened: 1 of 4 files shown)


BarbourSmith:
I just had this trigger during the calibration process when measuring the distance between the motors...which is awesome because we are stalling the motor for a fraction of a second there...but also an issue. 

Would it be safe to set the threshold to something large like 200mm when entering that step in the settings, and then reset it to normal when moving on?

blurfl:
Sure. Do we need to save the previous value to restore it?

BarbourSmith:
I think I can just send a hard coded value, and then pull the original value from the config file, right?

blurfl:
That's right, sending $42=200 as a macro doesn't change the setting in groundcontrol.ini, just in the EEPROM.

cnc_ctrl_v1/Report.cpp
```diff
@@ -236,6 +243,12 @@ void  returnError(){
         Serial.print(',');
         Serial.print(incSerialBuffer.spaceAvailable());
         Serial.println(F("]"));
+        float positionErrorLimit = 1.5;
```
BarbourSmith: @blurfl Is the setting for adjusting the positionErrorLimit being applied? I was just messing with it and it doesn't seem to change the threshold when I change the setting, is this hard coded value overridden?

blurfl:
This affects 'Test Motors/Encoders/ as well...

blurfl:
Not sure how to keep the  B04 'Test Motors/Encoders' from triggering the position error during the test. It of course will trigger it once the test is complete.
 Should we use a global flag in the firmware to suspend testing for the the error limit while the B04 is under way, or have B04 stash the value and restore it after the last axis has finished?

davidelang:
On Wed, 18 Apr 2018, Scott Smith wrote:

> Not sure how to keep the  B04 'Test Motors/Encoders' from triggering the position error during the test. It of course will trigger it once the test is complete.
> Should we use a global flag in the firmware to suspend testing for the the error limit while the B04 is under way, or have B04 stash the value and restore it after the last axis has finished?

I would take the attitude that B commands have no safety net, you do _exactly_ 
what they tell you to do. G commands should have the safety net of limits, stall 
detection, etc. But let the B commands remain unprotected so that they can be 
used for unexpected things.

BarbourSmith:
Great point @blurfl .

What if we just did a reload position at the end of the test so that wherever the sled is is where it's supposed to be at that point? Maybe we should have been doing that anyway?

blurfl:
> position reload

A great idea. Can you point at a place that shows how this is done?  I had a quick look but didn't spot it. I'm deep in a trig. rabbit hole over in gcodeCanvas, wanted to stay focussed there.

Here's a branch of firmware that implements a system flag to suspend evaluating position error, and uses it during the B04 process. 
https://github.com/blurfl/Firmware/tree/test-motors-needs-to-be-exempt-from-position-error-limit

BarbourSmith:
I believe that line **249** in Gcode.cpp is doing what we want to do:

`leftAxis.set(leftAxis.read());`

It just reads the current position and tells the axis that it is right where it should be

blurfl:
Great! Thanks for the quick answer💯👍. A PR is opened 😁
Collaboration is awesome!

BarbourSmith:
It is!!!!

PR #430, opened 2018-04-17
Replace G2/G3 line with equivalent G1 line

blurfl:
The problem here is floating point precision, but it can be identified by looking at the height of the chord described by the G2/G3 command. Problematic commands have a chord that resolves to 0.01mm or less. This code trapps that situation and calls G1 instead using the endpoint. This turns an arc with less than 0.01mm of bow into a straight line.
 This is discussed in issue #426, and applies to the issue addressed in PR#428, replacing the code from that PR.

### DIFF
--- a/cnc_ctrl_v1/Motion.cpp
+++ b/cnc_ctrl_v1/Motion.cpp
@@ -263,7 +263,7 @@ int   arc(const float& X1, const float& Y1, const float& X2, const float& Y2, co
             theta -= 2*pi;
         }
     }
-    if ((sign(theta) != sign(direction)) || (abs(chordHeight) < .01)) {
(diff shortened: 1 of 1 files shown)


BarbourSmith:
You are on 🔥 Way to take what was looking like an issue that would fall by the wayside as not frequent enough and tackle it completely!

blurfl:
"But wait, there's more..." Do look at the most recent comment in Issue #426 ...

BarbourSmith:
Hahha there always is! Do you want to add more changes in a new pull request or should I have left this one open?

blurfl:
The flyspeck I mentioned is in both GC and the firmware. I guess we could treat it as a separate issue. I'll open one, which place to choose?

BarbourSmith:
We could open issues in both places if we wanted, that way if we fix it in one place but not the other we would be able to note it

blurfl:
Ok, I'll do that.

PR #433, opened 2018-04-19
Test motors needs to be exempt from position error limit

blurfl:
Create a system flag to suspend checking for position error.
 Use it during B04 'Test Motors/Encoders'.
 Update our position after the test before re-enabling position error checking.


BarbourSmith:
Looks good! 👍 👍 

Way to pin down that corner case! 🎉 

Was there an issue where it was still triggering when doing the `rightAxis.set(rightAxis.read());`?

You are seriously on fire this week! 1.12 is going to be all you!

BarbourSmith:
Never mind I think my question didn't make sense because I misunderstood something 🙃

PR #434, opened 2018-04-19
Use pullups to check board version, use #defines for version

blurfl:
Use pullups to check board version
use #defines for version
add two more version pins for future use, pull them up as well

### DIFF
--- a/cnc_ctrl_v1/Config.h
+++ b/cnc_ctrl_v1/Config.h
@@ -35,6 +35,14 @@
 
 #define LOOPINTERVAL 10000 // What is the frequency of the PID loop in microseconds
 
+// Define version detect pins
+#define VERS1 22
+#define VERS2 23
+#define VERS3 24
+#define VERS4 25
+#define VERS5 26
+#define VERS6 27
+
 // Serial variables
 #define INCBUFFERLENGTH 128 // The number of bytes(characters) allocated to the
                             // incoming buffer.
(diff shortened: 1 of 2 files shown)


BarbourSmith:
This looks great!

I will test against all the board version in the morning and merge 🎉

BarbourSmith:
Hmmm this seems to cause issues on 1.2 although I'm seeing version 1.2 be detected:

image

Any idea what might be going on or what else I can test?

BarbourSmith:
Hang on, this looks to be an error on my part in the testing

blurfl:
I must be dense, what is the issue? By all means use a different approach, this is just a quick stab at it. 🤗

BarbourSmith:
> I must be dense, what is the issue? 

One of us is, but it's not you!

Ive been messing with the calibration stuff on my computer and I had it set up as top feed instead of bottom feed so everything was backwards 🙄 

Just tested again and all is well!! 👍 👍 🎉

blurfl:
That’s good! I think I should leave preparing for the merging of the TLE5206 board to you, if that’s ok?

BarbourSmith:
Do you mean moving the TLE5206 board into the /electronics repo?

blurfl:
I meant firmware. While my PR will recognize the TLT5206 board, it does not add the defines needed to make it run. I figured that just doing the version part would be an appropriate 'baby step'.

BarbourSmith:
OK, got it!

I'm making some progress on getting them made. They are a little weirded out by the whole we don't want any money thing but I think they will get it 😀

blurfl:
>They are a little weirded out by the whole we don't want any money thing

I think you could sell them if you wanted, they're shared in the same way as the boards you currently sell, aren't they?

BarbourSmith:
Its true and we will if we need to, but I think that if the factory can ship them direct it will be cheaper for everyone and they will be available sooner. We're really not that great at logistics 😀

PR #439, opened 2018-04-24
Added support for active low spindle automation.

### LINKED ISSUE #436: Add support for configuring spindle control signal high/low.
Currently I am setting up my spindle control logic to connect to a super-pid spindle controller.  It uses an active low to trigger it, and I notice in code that there is almost support to have option to have active low or high but it is hardcoded to be `boolean activeHigh = true;`   Enhancement would be to add ability to pull this value from sysSettings  , and a '$' command to set it.

Thanks.

CptanPanic:
I change the spindleAutomate settings to be an enum, so now you can be values below.  Note this will require a change in ground control, as it setup as a bool currently.  

```
enum SpindleAutomationType {
  NONE,
  SERVO,
  RELAY_ACTIVE_HIGH,
  RELAY_ACTIVE_LOW };
```

closes #436

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -263,9 +263,6 @@ void   Axis::test(){
         if (sys.stop){return;}
     }
     
-    Serial.print(F("Pos Change:  "));
-    Serial.print(encoderPos - motorGearboxEncoder.encoder.read());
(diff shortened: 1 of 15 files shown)


blurfl:
How about removing the delay from the ‘None’ code? No reason to wait around.

PR #444, opened 2018-05-03
Make it possible to measure either chain

BarbourSmith:
This pull request adds the ability to measure the distance between the motors by using either chain.

It also cleans up the directWrite function to call the regular write function to ensure consistency across PCB versions. This has been tested but it's a change which effects multiple other places in the code so it's worth mentioning.

This pull request is part of an effort to create an automated system for measuring differences in the left vs right chains

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -232,7 +232,12 @@ byte  executeBcodeLine(const String& gcodeLine){
     if(gcodeLine.substring(0, 3) == "B10"){
         //measure the left axis chain length
         Serial.print(F("[Measure: "));
-        Serial.print(leftAxis.read());
+        if (gcodeLine.indexOf('L') != -1){
+            Serial.print(leftAxis.read());
+        }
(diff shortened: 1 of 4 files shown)


davidelang:
can we kill these robot messages, let them do the work, but the message adds no 
value

and we don't vote on accepting pull requests, we discuss it, but we don't vote.

davidelang:
On Thu, 3 May 2018, BarbourSmith wrote:

> We do vote! You vote by giving that comment a thumbs up or a thumbs down which 
> lets everyone decide what is incorporated into the software and what isn't. If 
> the comment gets more 👎 than 👍 then the pull request is rejected and closed.
>
> It's a way to give everyone write access without giving anyone write 
> access...and I'm hoping that one of these days it will let me take a week off

Do people actually vote? Is there a way to vote via e-mail?

BarbourSmith:
There hasn't been much voting yet, but I'm hoping to start relying on the system entirely to merge pull requests so we can start testing it.

I don't think it is possible to vote by email right now

PR #445, opened 2018-05-08
Change folder name

BarbourSmith:
I think that the name for the folder containing the firmware as cnc_ctrl_v1 is a little confusing and that Maslow_Firmware might be a better name...unfortuately this will disrupt any branches whihc are out there and haven't been merged


blurfl:
This will probably get buried in the noise, but I think that the src_dir in platformio.ini needs to be changed to point to the new directory name, yes?

PR #446, opened 2018-05-11
Reword message about arcs

BarbourSmith:
It seems like this message is giving folks the impression that something has gone wrong when really the machine is being smart and preventing an issue.

I think it makes sense to take out the words oops and error which make perfect sense when you understand the code but which are intimidating if you don't know what is going on


blurfl:
Maybe "large-radius arc" instead of "large arc"?

BarbourSmith:
I agree. Done!

PR #448, opened 2018-05-11
Move bracket in alarm code

BarbourSmith:
When I was testing the reworded popup I noticed that the alarm triggered in every direction except when the right motor was moving to the left and I think this bracket needs to be moved over one place for the abs() to apply to the error and not the bool compare value 🙄


CptanPanic:
Good catch, those are tough ones to find.

PR #449, opened 2018-05-11
Tle5206 release

blurfl:
Merge the TLE5206 fork with the main. Most differences are found in System.cpp where board ID and pin assignments occur, and Motor.cpp where the pins get used.


Bee91303:
This works on the board I built out - Thru hole, passes motor tests.  I have 2 motors and moved them around the board.  Maslow had an issue getting Z kits out and have since fixed it so I will have a Z motor soonish.

PR #451, opened 2018-05-28
Convert arc with radius greater thanb 2500mm to a straight line

blurfl:
PR#420 added code to convert problematic G2/G3 arcs to G2 lines. This adds arcs with radius greater than 2500mm to the list of items converted. 

@Bar suggested 'a few thousand inches' [in Issue #426 ](https://github.com/MaslowCNC/Firmware/issues/426#issuecomment-379930897), this is less than that.

This addresses the issue discussed in [this forum topic](https://forums.maslowcnc.com/t/maslow-with-a-mind-of-its-own/4088).

### DIFF
--- a/cnc_ctrl_v1/Motion.cpp
+++ b/cnc_ctrl_v1/Motion.cpp
@@ -263,7 +263,7 @@ int   arc(const float& X1, const float& Y1, const float& X2, const float& Y2, co
             theta -= 2*pi;
         }
     }
-    if ((sign(theta) != sign(direction)) || ((abs(chordHeight) < .01) && (abs(theta) < 0.5))) {
+    if ((sign(theta) != sign(direction)) || ((abs(chordHeight) < .01) && (abs(theta) < 0.5)) || (radius > 2500)) {
       // There is a parameter error in this line of gcode, either in the size of the angle calculated
       //  or the chord height of the arc between the starting and ending points
       // In either case, the gcode cut was essentially a straight line, so 


bar:
@blurfl I think you might be confused since I did not suggest anything like that :)

blurfl:
Sorry for the incorrect reference, I've updated the comment to correct it. Thanks for your patience.

cnc_ctrl_v1/Motion.cpp
```diff
@@ -263,7 +263,7 @@ int   arc(const float& X1, const float& Y1, const float& X2, const float& Y2, co
             theta -= 2*pi;
         }
     }
-    if ((sign(theta) != sign(direction)) || ((abs(chordHeight) < .01) && (abs(theta) < 0.5))) {
+    if ((sign(theta) != sign(direction)) || ((abs(chordHeight) < .01) && (abs(theta) < 0.5)) || (radius > 25400)) {
```
BarbourSmith: How is the number 25400 computed? Not asking because I think there is anything wrong with it, just curious

blurfl: That’s 1000 inched turned into millimeters 😀

BarbourSmith: Got it 😀 Thanks

davidelang:
where is the limit that triggers the math error? how many digits of precision 
(and should we change the variables to double-precision floats to give more 
usable significant digits)

blurfl:
```
where is the limit that triggers the math error?
```
in Issue #426 you suggested making some sample g-code files with large radius curves to see if they have problems. Have you had a chance to investigate that? The results would probably be helpful in this case.

blurfl:
```
should we change the variables to double-precision floats to give more usable significant digits
```

Unfortunately, on the Uno and other ATMEGA based boards, doubles occupy 4 bytes. That is, the double implementation is exactly the same as the float, with no gain in precision.

davidelang:
There isn't an 8 byte float available? not even with an alternate math library?

what about going to fixed-precision math?

BarbourSmith:
My instinct is that when we are doing sin and cos on arcs with a radius of a mile we're never going to be able to compute that result accurate to within a mm. With arcs so large the right behavior is to treat the segment as a straight line and compute it as such. If the amount of the arc is much less than what we can measure it is safe to treat it as a straight line just like @blurfl has made it

beretta42:
Sorry to drag this issue/commit up again.  1000 inches does not make a mile :)  I did a back-of-envelope calculation.. a circle with radius 1000 inches (25400mm)  yields an arc height of approximately 29mm.  This seems far from a straight line.

blurfl:
```
1000 inches does not make a mile
```
That's true, it's only 83 feet. With that radius the arc across 8 feet would be 1.15 inches high. The 1000 inch value was picked out of the air, tested and found to correct the issue being addressed by a fair margin. The troublesome arcs in the forum discussion were in the quarter- to half-mile range, with mantissas out to four places.
 In post-processing a complex shape, using different resolution settings, using an arc height less than that or converting the arc to smaller straight lines would create a file that would cut without the straight-line approximation. 
 We're bumping our noses on the limits of the floating point implementation versus various methods used by assorted post processors to approximate complex curves. Add to that the time and processing power limits of a low-cost microprocessor doing interrupt-driven PID control on a machine with six interrupt sources - there are many opportunities for contribution! 😁

davidelang:
so, a floating point number precision is 24 bits, or just over 7 decimal digits. 
( https://en.wikipedia.org/wiki/Single-precision_floating-point_format )

so if we are tracking things to 0.001mm, that means that around a radius of 
10,000mm things break

unfortunantly, according to http://www.handymath.com/cgi-bin/arc18.cgi this 
still has an arc height of 74mm over a 8' wide cut.

you have to go to a radius of 1,000,000 to get the arch height down to <1mm, 
about 100x what we can do with floating point.

Now, if instead of tracking down to 0.001mm, we only track down to 0.1mm, then a 
limit of 10,000mm for the radius would get the arc height that we throw away 
down to <1mm, but I think that only tracking positions down to 0.1mm is going to 
end up causing problems elsewhere.

I think that we are going to end up having to bite the bullet and move to fixed 
point math for this to survive the single-precision limit (and that will wreak 
havoc with the trig functions)

David Lang

davidelang:
As I just noted ( I think in a different thread, apologies if this is a dup 
here)

if we are trying to track things to 0.001mm, then we run out of precision on the 
float around 10,000mm, and at that point the arc that we would have is 74mm 
high.

blurfl:
```
move to fixed point math
```
AVR fixed point is pretty rare, Arduino compatible implementations are very rare. Do you have one in mind?

davidelang:
On Fri, 8 Jun 2018, Scott Smith wrote:

> ```
> move to fixed point math
> ```
> AVR fixed point is pretty rare, Arduino compatible implementations are very rare. Do you have one in mind?

no, I am the type that would probably just open code some functions to do it.

PR #454, opened 2018-06-12
Guestbook - Ned Pull Request test

neilvangeffen:
Put my name in the guest book. Time to do a pull request and see if it all worked out ok!

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -252,50 +252,8 @@ void  Kinematics::xyInverse(float xTarget,float yTarget, float* aChainLength, fl
     //Confirm that the coordinates are on the wood
     _verifyValidTarget(&xTarget, &yTarget);
 
-    //Set up variables
-    float Chain1Angle = 0;
-    float Chain2Angle = 0;
-    float Chain1AroundSprocket = 0;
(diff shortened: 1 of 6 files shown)


blurfl:
First, welcome @neilvangeffen ! Unfortunately, it looks like the test has gone awry - beyond just entering into the Guestbook, it would make many fundamental changes to the master firmware. I'm sure that's not what you wanted, so I'm going to close it rather than let it stand.

 Are you experimenting with an XY version? You will want to keep that code in a local branch separate from the master branch. As it stands right now, the master firmware for everyone would be changed to XY by this pull request.

 I would suggest making a branch for 'sign-the-guestbook', and a separate one for 'investigate-XY-operation'. Then create a pull request for the guestbook one, but _not_ for the branch you'll use for your local development. A pull request is only needed when you want to propose changing the firmware for everyone to use what you've written.

BarbourSmith:
Very close though! It looks like the PR to sign the guest book wen't perfectly, all that you want to change for the next one is to make a new branch before working on the XY code. 

Thanks for using the guest book! It's been neglected recently and it's good to see it in use again. 👍 👍

neilvangeffen:
Cheers guys. Will try again :-) I thought pull request was for just a specific commit, not a whole branch.

So now i know and will get it right next time. Cheers!

PR #460, opened 2018-08-16
Update INSTRUCTIONS.md

ted1735:
the link does not show up in the community garden. not sure why but it may be the period at the end of the sentence just deleted the period so it doesn't interfere with link...


blurfl:
Good catch, and a welcome contribution!
 Removing the period probably won't have the desired effect, though. Take a look at the link in Step two which is properly formatted as [description](link).
 How about altering this PR to implement that format?

blurfl:
Sorry, my comment was poorly structured and got rendered instead of showing what I meant. A link should have two parts, a description enclosed in square brackets followed by the link enclosed in parentheses, like so:

```
[https://github.com/MaslowCNC/Firmware/releases](https://github.com/MaslowCNC/Firmware/release)
```
Let's keep the https link as more secure...

PR #468, opened 2018-10-16
Adjust encoderStep defaults

johnboiles:
@madgrizzle will have to chime in on the justification for this change. I just pulled it from his development branch!

### DIFF
--- a/cnc_ctrl_v1/MotorGearboxEncoder.cpp
+++ b/cnc_ctrl_v1/MotorGearboxEncoder.cpp
@@ -125,7 +125,7 @@ void MotorGearboxEncoder::setEncoderResolution(float resolution){
     
     */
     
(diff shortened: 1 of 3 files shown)


cnc_ctrl_v1/MotorGearboxEncoder.cpp
```diff
@@ -125,7 +125,7 @@ void MotorGearboxEncoder::setEncoderResolution(float resolution){
     
     */
     
-    _encoderStepsToRPMScaleFactor = 60000000.0/resolution; //6*10^7 us per minute divided by 8113.7 steps per revolution
+    _encoderStepsToRPMScaleFactor = 60000000.0/resolution; //6*10^7 us per minute divided by 8113.73 steps per revolution
```
BarbourSmith: Thanks for taking the time to update the comment too

PR #471, opened 2018-10-23
Simplified Chain Tolerance Correction

johnboiles:
This is another bit of code I found in @madgrizzle's development branch. I believe this to be a solid simplification of how the chain tolerance correction works.

In the current system, GroundControl accepts a positive or negative percentage adjustment to adjust the chain tolerance for each chain (e.g. 1.3% if your chain is 1.3% longer than expected), it then turns that into a calculated value for `distPerRotLeftChainTolerance` by multiplying the adjustment by `distPerRot`. That value is then sent to the firmware, where it is divided by `2*Pi` and stored as `RleftChainTolerance`. It's then used in place of the sprocket radius `R` in the firmware for all calculations that involve the sprocket's radius.

With this change, GroundControl continues to accept a positive or negative percentage adjustment for each chain (current settings will continue to work), but does no calculation on that value. Then that adjustment is sent to the firmware as `leftChainTolerance`. That tolerance value is simply multiplied by the calculated chain length to derive the adjusted chain length.

Additionally, I think this is maybe _slightly_ more accurate. The current chain tolerance adjustment adjusts for the bit of chain that is still on the sprocket (`Chain1AroundSprocket`). I _think_ the length of the chain around the sprocket is a function of the radius of the sprocket, and not the tolerance in the chain, though it's likely not a big enough value to matter one way or another. It's trivial to make this change account for that as well.

The corresponding GroundControl change is MaslowCNC/GroundControl#783
Discussion in the forums is here: https://forums.maslowcnc.com/t/simplified-chain-tolerance-correction-pull-request/7395

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -59,6 +59,8 @@ void Kinematics::recomputeGeometry(){
     _xCordOfMotor = sysSettings.distBetweenMotors/2;
     _yCordOfMotor = halfHeight + sysSettings.motorOffsetY;
 
+    leftChainTolerance = sysSettings.distPerRot/sysSettings.distPerRotLeftChainTolerance;
+    rightChainTolerance = sysSettings.distPerRot/sysSettings.distPerRotRightChainTolerance;
 }
 
 void  Kinematics::inverse(float xTarget,float yTarget, float* aChainLength, float* bChainLength){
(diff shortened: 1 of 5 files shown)


BarbourSmith:
Great! I'm about to run a longer cut and give this a solid test. Excellent picking!

johnboiles:
I'm glad you were able to take a look and I'm glad you'll be able to test! Thanks @BarbourSmith

BarbourSmith:
I am two hours into a test with no issues so far. Gets my 👍 for sure. Great work!

We should keep in mind that this change requires updating Ground Control to work properly so I bet we will see a couple questions in the forums triggered by that.

johnboiles:
Great! What are you cutting?

BarbourSmith:
I'm making a skateboard rack for a friend:

image

johnboiles:
Neat! Looking good. Thanks for sharing

PR #474, opened 2018-11-03
Experimental optical calibration support

johnboiles:
Just wanted to put this out there for discussion and feedback. This is @madgrizzle's optical calibration correction feature from his fork. The two of us have been using it with success. It's caused my cuts to be more accurate than before.

There are two ways optical calibration corrections can work.

1) **Interpolation mode** A correction matrix from 465 calibration points (31 columns, 15 rows) is uploaded into EEPROM. Moves are adjusted by interpolating the error between the points to find the positional error at any position on the board.

2) **Curve fit mode** Curve fit coefficients are calculated by GroundControl based on the results of the calibration. The coefficients are stored into EEPROM. Moves are adjusted by using the curve fit equations to find the positional error at any position on the board.

Forum discussion is [here](https://forums.maslowcnc.com/t/experimental-optical-calibration-pull-request/7557). 

Original Optical Calibration thread is [here](https://forums.maslowcnc.com/t/optical-calibration-demo-and-three-hours-working-on-a-bug/6733).

### DIFF
--- a/cnc_ctrl_v1/Calibration.cpp
+++ b/cnc_ctrl_v1/Calibration.cpp
@@ -1,0 +1,59 @@
+/*This file is part of the Maslow Control Software.
+
+The Maslow Control Software is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Maslow Control Software is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
+
+Copyright 2014-2017 Bar Smith*/
+
+// This file contains the machine settings that are saved to eeprom
+
+// EEPROM addresses 300 and up can be used by Maslow.  Under 300 was used
+// previously by pre v1.00 Firmware.
+
+#include "Maslow.h"
+#include <EEPROM.h>
+
+void initializeCalibration(){
+    /*
+    initializes the data to zero
+    */
+    Serial.print(F(" Zeroing Calibration\r\n"));
+    for (int x=0; x<31; x++){
+      for (int y=0; y<15; y++){
+        calibration.xError[x][y]=0;
+        calibration.yError[x][y]=0;
+      }
+    }
+}
+
+byte calibrationUpdateMatrix(int _x, int _y, int xValue, int yValue) {
+  if ( (_x==31) && (_y==15) )
+  {
+    settingsSaveToEEprom();
+    return(STATUS_OK);
+  }
+  else
+  {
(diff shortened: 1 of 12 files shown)


cnc_ctrl_v1/Calibration.cpp
```diff
@@ -0,0 +1,46 @@
...
+  } else {
+    if (((x>=0) && (x<31)) && ((y>=0) && (y<15))) {
+      calibration.xError[x][y] = xValue;
+      calibration.yError[x][y] = yValue;
+      if ((x==15) && (y==7)) {
+        kinematics.init();
+      }
+      return(STATUS_OK);
+    }
+    return(STATUS_INVALID_STATEMENT);
```
johnboiles: Might be simpler to write this as
```C
byte calibrationUpdateMatrix(int x, int y, int xValue, int yValue) {
  if ((x == 31) && (y == 15)) {
    settingsSaveToEEprom();
    return(STATUS_OK);
  } else if ((x < 0) || (x >= 31) || (y < 0) && (y >= 15)) {
    return(STATUS_INVALID_STATEMENT);
  }
  calibration.xError[x][y] = xValue;
  calibration.yError[x][y] = yValue;
  if ((x==15) && (y==7)) {
    kinematics.init();
  }
  return(STATUS_OK);
}
```

cnc_ctrl_v1/Calibration.cpp
```diff
@@ -0,0 +1,46 @@
...
+along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
+
+Copyright 2014-2017 Bar Smith*/
+
+#include "Maslow.h"
+#include <EEPROM.h>
+
+void initializeCalibration() {
+  Serial.print(F("Zeroing Calibration\r\n"));
+  for (int x=0; x<31; x++) {
```
johnboiles: Probably would be good to pull 31 and 15 into constants.

cnc_ctrl_v1/Calibration.cpp
schmittjoshc: I wasn't able to go through and confirm anything in detail.  We might call it enableXYAdjustment, rather than enableOpticalCalibration.  The (X,Y) adjustments could be determined via a bunch of mechanisms besides optical calibration.  In triangularInverse, it looks like you verifyValidTarget twice: once before adjustment and once after.  Is that necessary?  Would this be better implemented as an if/else, where the first verifyValidTarget is in moved into the else block?

BarbourSmith:
I feel like until we have a number of folks using it that the added complexity isnt worth it just yet. I want to keep an eye on feature creep and only add things with wide spread appeal even when they are awesome features like this. It seems like without a place to get the calibration pattern and the camera the feature probably won't see significant use. Now if someone selling kits started shipping them with the hardware I would vote to add this right away

cnc_ctrl_v1/Calibration.cpp
johnboiles: The naming change is a good idea. 

For `verifyValidTarget` it’s a good question. I’ll take a look again and see if this can get cleaned up.

johnboiles:
@BarbourSmith I agree (I downvoted). My goal here was to get some other eyes on it to see if anyone has ideas to improve it. Thanks for taking a look.

PR #479, opened 2019-01-21
Change operation associated with ChainTolerance

madgrizzle:
This was discovered by @c0depr1sm and @joshua in this thread: https://forums.maslowcnc.com/t/holey-triangular-calibration/6240/191.
The original implementation of ChainTolerance modified the sprocket's mmPerRotation (increasing it for a positive ChainTolerance value).  The effect was to reduce the amount of  chain fed out.  This was adjusted in a previous PR where the ChainTolerance only modified the ChainStraight length in triangularInverse (because there shouldn't be any chain stretch associated with the wrap around the sprocket).  However, the formula used actually increases the amount of chain that is fed out (i.e., multiplying by a factor of 1+ChainTolerance) instead of decreasing it.  So, this PR changes the operator such that the ChainTolerance is divided into the ChainStraight rather than multiplies it resulting in a smaller chain length amount being fed out.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -236,8 +236,8 @@ void  Kinematics::triangularInverse(float xTarget,float yTarget, float* aChainLe
     Chain2Straight *= (1 + ((sysSettings.chainSagCorrection / 1000000000000) * pow(cos(Chain2Angle),2) * pow(Chain2Straight,2) * pow((tan(Chain1Angle) * cos(Chain2Angle)) + sin(Chain2Angle),2)));
 
     //Calculate total chain lengths accounting for sprocket geometry and chain sag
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Excellent catch! 👍 👍 ❤️ ❤️ 

It was my experience that messing with the chain tolerance generally decreased my accuracy, I wonder if this could be the reason. Has anyone done a comparative calibration test yet?

madgrizzle:
This was a recent error.  I recall you having that issue when using your original implementation which I think didn't have this problem.

c0depr1sm:
Actually, I did somewhat of comparative test: The Zipper Tree Chalenge was done with an [alternate triangular inverse function](https://github.com/c0depr1sm/Firmware/blob/8e622d8f108facdd577a045094471593c8554d55/cnc_ctrl_v1/Kinematics.cpp#L289)... including **that chain tolerance calculation fix**.

BarbourSmith:
And the results speak for themselves. Very impressive!

PR #480, opened 2019-02-15
Change originalChainLength to 1651 mm

madgrizzle:
Though it really doesn't matter to do this here, because ground control overwrites the default values, this proposal changes the originalChainLength default value to 1651 mm, which is an integral of 6.35 mm (the pitch of a #25 chain) and therefore would result in a tooth being at 12 o'clock after the chains are extended.  Using this value can allow someone to manually reset the chains without having to go through the long, tedious process of having the motor feed out the chain.

### DIFF
--- a/cnc_ctrl_v1/Settings.cpp
+++ b/cnc_ctrl_v1/Settings.cpp
@@ -74,7 +74,7 @@ void settingsReset() {
     sysSettings.rotationDiskRadius = 250.0;  // float rotationDiskRadius;
     sysSettings.axisDetachTime = 2000;   // int axisDetachTime;
     sysSettings.chainLength = 3360;   // int maximum length of chain;
-    sysSettings.originalChainLength = 1650;   // int originalChainLength;
+    sysSettings.originalChainLength = 1651;   // int originalChainLength;
     sysSettings.encoderSteps = 8113.73; // float encoderSteps;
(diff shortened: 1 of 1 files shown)


davidelang:
can we add a warning in GC if the originalChainLength is not a multiple of the 
link size?

David Lang

BarbourSmith:
I feel like a warning might scare more people than it helps since most
people won't notice the difference.

davidelang:
On Fri, 15 Feb 2019, BarbourSmith wrote:

> I feel like a warning might scare more people than it helps since most
> people won't notice the difference.

this error creates about 2mm of cut error after the first time someone resets 
the chain.

I would unconditionally change saved values from 1650 to 1651, but for any other 
value, I would warn about it and let the user override the warning. We could 
limit the warning to a one-time thing or a check when someone goes to change 
settings.

David Lang

GeroBH:
Why is there a warning from "Travis CI" about a missing 'Servo.h'?

Off-topic! How they build a global team and the tools used is worth a watch, as the downsides are also explained. https://www.youtube.com/watch?v=XAU5q-biY28

From the quick view it looks like a company supporting in software development?
Who hired them to follow the Maslow Firmware and complain about a file found in the Aduino IDE in /home/{username}/arduino-1.8.3/libraries/Servo/src?

PR #481, opened 2019-02-17
Chain length calculation improvement: stretch compensation 

c0depr1sm:
Following tests with the help of [beta testers on the forum](https://forums.maslowcnc.com/t/call-for-beta-testers-chain-stretch-compensation-improvement/8742), proposed improved firmware is ready for a pull request:
 + It implements chain tension and stretch calculations to improve chain length calculation.
 + It uses a default sled weight of [25 lbs](https://forums.maslowcnc.com/t/bricks-on-sled/7235) (11.6 kilo) that you can set with macro $46=value where value  =kilo * 9.8 or lbs * 2.2*9.8. That settings would be eventually added to GroundControl.

The community will also enjoy **improved standard** Maslow Calibration results because:
 + the chain stretch compensation reduces the motor distance estimation error
 + the sag correction factor is still in use with that firmware.

Although the Standard calibration results will be improved, you might want to keep your known values if you trust them to be better.

### Why that step?
What are the benefits of making this firmware improvement in such a step?
 + More constant vertical scale accross the workspace.  
 + Users of the triangular link should get better results out of the current GroundControl Calibration scheme.    
 + When Ground Control will be improved with sled weight setting parameter, it will be easy to set it specifically for their sled.
 + With that update, one can also start experimenting with chain tolerance optimisation using the [Holey triangular Calibration (instructions here)](https://forums.maslowcnc.com/t/holey-triangular-calibration/6240/176?u=c0depr1sm). That without installing another firmware: Just download in a separate folder the HoleyCalibration GroundControl, then you'll find there the gcode and python scripts. 
 + (EDIT) It gets users ready for getting a better GroundControl with sled weight setting and is in line with Holey Triangular Calibration's features roadmap.

### DIFF
--- a/cnc_ctrl_v1/Kinematics.cpp
+++ b/cnc_ctrl_v1/Kinematics.cpp
@@ -73,15 +73,15 @@ void  Kinematics::inverse(float xTarget,float yTarget, float* aChainLength, floa
     _verifyValidTarget(&xTarget, &yTarget);
     
     if(sysSettings.kinematicsType == 1){
-        quadrilateralInverse(xTarget, yTarget, aChainLength, bChainLength);
+        _quadrilateralInverse(xTarget, yTarget, aChainLength, bChainLength);
     }
     else{
-        triangularInverse(xTarget, yTarget, aChainLength, bChainLength);
+        _triangularInverse(xTarget, yTarget, aChainLength, bChainLength);
     }
     
 }
 
-void  Kinematics::quadrilateralInverse(float xTarget,float yTarget, float* aChainLength, float* bChainLength){
+void  Kinematics::_quadrilateralInverse(float xTarget,float yTarget, float* aChainLength, float* bChainLength){
 
     //coordinate shift to put (0,0) in the center of the plywood from the left sprocket
     y = (halfHeight) + sysSettings.motorOffsetY  - yTarget;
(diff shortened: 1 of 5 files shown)


c0depr1sm:
Looks like the automated travis builder does not respond like Arduino's IDE: there seems to be missing a Servo.h in the travis automated builder files? That is not a new file with this pull request. It looks like a file that was called for compilation since encoders were used?

GeroBH:
> Looks like the automated travis builder does not respond like Arduino's IDE: there seems to be missing a Servo.h in the travis automated builder files? That is not a new file with this pull request. It looks like a file that was called for compilation since encoders were used?

I've sent them this email:
Dear Support, (Antwort gerne auch auf Deutsch)

With 2 pull requests we added to:
https://github.com/MaslowCNC/Firmware/pull/480
and
https://github.com/MaslowCNC/Firmware/pull//481
there is a warning that all tests have failed with information leading here:
https://travis-ci.org/MaslowCNC/Firmware/builds/494405987?utm_source=github_status&utm_medium=notification
and here
https://travis-ci.org/MaslowCNC/Firmware/builds/493896920?utm_source=github_status&utm_medium=notification

Looking at the error, it's complaining about a Arduino library that can be found in the local installation of the IDE here:
/home/{username}/arduino-1.8.3/libraries/Servo/src/Servo.h

To me this looks more like an issue with your PlatformIO then with our firmware that compiles and runs local without issues.
Hope you could find this information useful and wondering who hired you to monitor our software. :-)

Hochachtungsvoll,
Gero Presser

schmittjoshc:
My only thing is: it looks like all the changes to the kinematics.cpp file are a direct copy-paste of the original ones I made in https://github.com/schmittjoshc/MaslowCNCFirmwareUpdate.  However, there are no links or references to those revisions.  The only exception is the implementation of chain-sag compensation.  You've chosen to implement the equations utilizing the quadratic assumption.  If you look at the equations, the full catenary equation is simpler and more computationally efficient; it is able to re-use the tension values from the tension calculations.

c0depr1sm:
You are right @schmittjoshc , I should have referenced you work in the comments. I will correct the references in my next addition to the firmware. About the chain sag calculation, I will open another topic on the forum to define the proper evolution and support.

PR #489, opened 2019-03-28
Fix to EEPROM write issue

madgrizzle:
This adds a flag (writeStepsToEEPROM) to the sys structure that is raised whenever an axis is attached.  The systemSaveAxesPosition() function tests that this flag is true and that all axes are detached.  If so, it calls settingsSaveStepstoEEprom() and after the EEPROM is saved, it clears the flag.  Therefore, the only time steps should be saved to EEPROM is after an axis becomes attached and then when all axes become detached.  So that's should only happen 2 seconds of idle after a motor move.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -197,6 +197,7 @@ int    Axis::detach(){
 
 int    Axis::attach(){
      motorGearboxEncoder.motor.attach();
+     sys.writeStepsToEEPROM = true;
      return 1;
 }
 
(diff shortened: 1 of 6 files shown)


BarbourSmith:
This looks very cleanly implemented. Excellent work!

I'm still pretty surprised that this issue exists. I know for a fact that we used to write to the eeprom only once two seconds after the motor was detached because I remember testing that with print statements early on, but the way the system works was refactored later and we could have changed the behavior.

I don't have access to an aduino right now to test, but can we confirm that a print statement added at line 163 in Settings.cpp `EEPROM.put(310, sysSteps);`, runs regularly?

If we are concerned about EEPROM fatigue we can move to a new location in memory from now on. I know we have a lot of extra slots although we would have to make it so the position was moved to the new location automatically so people won't be bothered by the change

madgrizzle:
I can try it out with fakeservo tommorow/this weekend.  If surprised it's writing, but I'm working on porting code to a teensy and in testing the motors, I ran into issues and the culprit was that it was writing to eeprom during the maslowDelay calls.. once I prevented it writing, my issues went away.

madgrizzle:
Oh, cool.. I updated the code in patch-4 and it added it to the PR.

PR #492, opened 2019-04-28
Line added to the guestbook

esspe2:
Added my name to the guestbook
and a slight correction of GCode.cpp
and a cancellation of this change ( I keep it for another PR)

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -268,7 +268,7 @@ byte  executeBcodeLine(const String& gcodeLine){
             execSystemRealtime();
             if (sys.stop){return STATUS_OK;}
         }
-        sys.state = (sys.state | (!STATE_POS_ERR_IGNORE));
+        sys.state = (sys.state & (~STATE_POS_ERR_IGNORE));
         return STATUS_OK;
     }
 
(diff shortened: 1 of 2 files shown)


blurfl:
Learning git is a task, for sure. 
One good practice is to put unrelated changes into separate PRs. For instance, the 'Corrected bit-masking...' would make sense as a separate PR.
 Separating changes makes testing the individual effects easier. It's not a big issue in this case, but you might try it for the practice?

esspe2:
Yes, I think like you, but I don't see a way to open a second PR, and all commits go in the open one.
Agreed, the next time I won't mess two unrelated changes

blurfl:
👍🏻 You're good. The guestbook is a way to get started learning about git.
 Look into the idea of 'git branch'. That's a way of gathering together commits to create a PR. Use separate branches for separate PRs. In this case, a branch for 'guestbook sign-in' and a separate branch for 'Corrected bit-masking...' would work.

BarbourSmith:
Excellent work!

Git can be pretty intimidating at first, but once you have it down the collaboration it allows is wonderful. There is a program called Github Desktop (https://desktop.github.com/) which can be easier to jump into.

Also as a quick introduction to how things work around here (which is a little different from other projects) this pull request will be automatically be merged by @MaslowCommunityGardenRobot in 48 hours as because @blurfl and I voted for it with the 👍

esspe2:
Thanks for all the support!
Happily there also is a Github Desktop for linux (https://github.com/shiftkey/desktop), I'll give it a try.

PR #493, opened 2019-04-28
Correction of some bitwise operations in GCode.cpp  

esspe2:
Convert direct Bitwise operations by macros defined in NutsAndBolts.h, less error-prone

### DIFF
--- a/guestBook.txt
+++ b/guestBook.txt
@@ -40,5 +40,3 @@ Tanju B - Hello! we finally making cuts down at the Maker Station makerspace in 
 Maslow Bahrain (GeroBH) made it into the guestBook ;-) 21 Feb 2018
 
 Morriz was here - 9/5/2018
-
-esspe2 was here, working on linux,  hi from Belgium ! 28 Apr 2019
--- a/guestBook.txt
+++ b/guestBook.txt
@@ -40,3 +40,5 @@ Tanju B - Hello! we finally making cuts down at the Maker Station makerspace in 
 Maslow Bahrain (GeroBH) made it into the guestBook ;-) 21 Feb 2018
 
 Morriz was here - 9/5/2018
+
+esspe2 was here, working on linux, hi from Belgium ! 28 Apr 2019
--- a/guestBook.txt
+++ b/guestBook.txt
@@ -41,4 +41,4 @@ Maslow Bahrain (GeroBH) made it into the guestBook ;-) 21 Feb 2018
 
 Morriz was here - 9/5/2018
 
-esspe2 was here, working on linux, hi from Belgium ! 28 Apr 2019
+esspe2 was here, working on linux,  hi from Belgium ! 28 Apr 2019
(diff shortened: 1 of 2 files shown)


guestBook.txt
```diff
@@ -40,3 +40,5 @@ Tanju B - Hello! we finally making cuts down at the Maker Station makerspace in
 Maslow Bahrain (GeroBH) made it into the guestBook ;-) 21 Feb 2018
 
 Morriz was here - 9/5/2018
+
+esspe2 was here, working on linux,  hi from Belgium ! 28 Apr 2019
```
BarbourSmith: It looks like the guest book change snuck in here too 😀

BarbourSmith:
I am not in a position right now to test this on hardware so I am not going to vote, but this generally looks good to me 👍

blurfl:
**I think this will need some testing. It compiles, but...**
 It looks to me like these functions from NutsAndBolts.h treat the second parameter as a mask instead of as a single bit number as the original code did. That clears, sets or test more bits that intended.
```
// These are nifty functions from Grbl
#define bit_true(x,mask) (x) |= (mask)
#define bit_false(x,mask) (x) &= ~(mask)
#define bit_istrue(x,mask) ((x & mask) != 0)
#define bit_isfalse(x,mask) ((x & mask) == 0)
```
The pre-defined Arduino functions [bitSet()](https://www.arduino.cc/reference/en/language/functions/bits-and-bytes/bitset/) and [bitClear()](https://www.arduino.cc/reference/en/language/functions/bits-and-bytes/bitclear/) work on single bits as the original code does.

 The logic test in the line 
```
if (line_flags & LINE_FLAG_COMMENT_PARENTHESES) { line_flags &= ~(LINE_FLAG_COMMENT_PARENTHESES); }
```
seems plain enough without changing it, though it might be re-written as 
```
if (line_flags & LINE_FLAG_COMMENT_PARENTHESES) { bitClear(line_flags, LINE_FLAG_COMMENT_PARENTHESES); }
```
to clarify the bit-clearing operation.

esspe2:
Ok, 
I understand: beware of the definition, bit number vs bitmask, good point!
By chance this flag is already defined as a bitmask so the macro still works here:  
GCode.h:
#define LINE_FLAG_COMMENT_PARENTHESES bit(0)

As for the readability, both ways seem valid, more a question of taste.
I thought better to use macros defined for those cases, which also permit to avoid confusions between  "!" and "~".

Well, in the last case the 'if' could even be dropped, since the only action is to clear the bit if it is set ;-)
What do you think? 

GCode.cpp:
bitClear(line_flags, LINE_FLAG_COMMENT_PARENTHESES);

aprospero:
esspe2 is right, the nutsandbolts macros from grbl use masks, the builtin arduino functions use bit numbers. Since LINE_FLAG_COMMENT_PARENTHESES is in fact a mask (which masks only one bit) bitClear can't be used in this context but bit_false is ok.

I'm with esspe2, the 'if' should be dropped. The compiler whould dop it anyway and the results is easier to read.

guestBook.txt
esspe2: Yeah, I have created the "correction" branch after the guest book change, so the modification was included, silly mistake ;-)

I'll see if I'm able to remove that file from this branch before the robot does the merge.
While I think that won't break anything.
And my other PR "Line added to the guestbook" should be merged before.
But some useful git practice btw.

blurfl:
You're both right, the changes work as expected.  👍

PR #494, opened 2019-05-02
detach() should run once per state change

blurfl:
No need to repeat the operation once it has been accomplished.

### DIFF
--- a/cnc_ctrl_v1/Motor.cpp
+++ b/cnc_ctrl_v1/Motor.cpp
@@ -77,7 +77,7 @@ void Motor::detach(){
         digitalWrite(_pin2,    LOW) ;
       }
     }
-    attachedState = 0;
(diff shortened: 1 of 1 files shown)


esspe2:
Positive vote for me; 
Btw, what happens to the pwm pin when the motor is detached ?
Since it is connected to L298's Enable pin, it could be pulled low.
I don't know if the pwm interferes, so the pwm value could also be dropped to 0 eg:
```
 } else {
    //stop the motor
    digitalWrite(_pin1,    LOW);
    digitalWrite(_pin2,    LOW);
    digitalWrite(_pwmPin,  LOW);   // Suggested modification
    write(0);                      // Suggested modification
  }
```

esspe2:
Positive vote for me, seems good practice.

PR #495, opened 2019-05-02
Update guestBook.txt and say Hello, again!

esspe2:
Line added to the Guestbook (again); 
since it had been deleted by my second PR, let's put the right content back after all.

This time I first have created a special branch (Gestbooks-revenge) before adding the line; 
hopefully it will be cleaner that way.


BarbourSmith:
Nice work with the branches 😀

PR #498, opened 2019-05-07
Move directWrite outside the loop

esspe2:
No need to repeat 1000 times the same actions, it seems sufficient to do the directWrite only once before the loop instead.


blurfl:
Glad to see you combing through the code @esspe2! Keep up the good work! 💯👍

esspe2:
Thanks for your good words!

The call chain to the eeprom write is more like this:
>      maslowDelay() 
>      -> execSystemRealtime() 
>      -> systemSaveAxesPosition() 
>      -> settingsSaveStepstoEEprom() 
>      -> EEPROM.put(310, sysSteps)
so this change doesn't really improve the eeprom issue, but I still feel it useful.

Issue #500, opened 2019-05-20
 'TestMotors' bug for TLE5206 boards


blurfl:
Somewhere after the ~v1.25~ _v1.26 release_ one of the changes has caused TLE5206 boards to fail the Axis::test() function. Stock Maslow boards are not affected. _The release v1.26 runs correctly._
 When the test is run, each motor runs and passes Direction 1 but does not run in Direction 2.

blurfl:
git bisect finds that PR #486 is the last version where the motor tests worked correctly, PR #489 marks the presence of an issue.

Digging deeper, [this commit](https://github.com/MaslowCNC/Firmware/pull/489/commits/2c41d174b273a5729d66d1540abdb30d141fcb1a) seems to be related to the problem. Before the change in this commit, the motor test runs normally on the stock Maslow v1.2 board as well as the TLE5206. After this change to the repo as it stood at that point, both boards fail to run properly but in a manner different from the behavior described in the initial issue. 
 I've tried backing out the later changes to Axis::test() and Motor:::detach() which seemed likely possibilities, no joy. I'll keep digging, suggestions welcome.

BarbourSmith:
Hmmm it's interesting that an issue like that would be caused by writing to the eeprom. I wouldnt expect that behavior.

I think this issue brings up a larger question about if there is a way to do hardware testing automatically on pull requests before we merge them. I will look into building some sort of rpi Arduino rig that will put some hardware through it's paces and comment on pull requests if all tests pass.

blurfl:
I don't think it has anything to do with writing to the EEPROM, but at that point in the change commit sequence the frequent writes affected something in the test timing that caused all boards to fail the tests.
After PR#495, only the TLE5206 boards fail, and they only fail the second half of each motor test. If you swap the forward and reverse cases within the test, it is still the second case that fails. And to make matters stranger, there's this: the two test cases are 
>motorGearboxEncoder.motor.directWrite(255);

and run for a while and check that the encoder counts went the direction expected then
>motorGearboxEncoder.motor.directWrite(-255);

and again check the encoder counts.

If you change the _second_ case to directWrite(-254) instead of -255 (or 254 if swapping the sequence), the TLE5206 will pass all the motor tests. _But why_? There has to be something in the TLE5206 section of Motor::write() but I haven't trapped it yet. Ideas welcome...

BarbourSmith:
[...]

PR #501, opened 2019-05-21
Avoid PWM value 255 for TLE5206

blurfl:
This patch only affects TLE5206 boards. If the PWM value is 255 it is changed to 254. This works around the  symptom noted in issue #500

### DIFF
--- a/cnc_ctrl_v1/Motor.cpp
+++ b/cnc_ctrl_v1/Motor.cpp
@@ -144,6 +144,7 @@ void Motor::write(int speed, bool force){
             }
         } 
         else { // TLE5206
+            speed = constrain(speed, 0, 254); // avoid issue when PWM value is 255
             if (forward) {
                 if (speed > 0) {
                     if (usePin2) {


blurfl:
I've tested this against Maslow boards v1.1, v1.2 and a TLE5206 board. All those pass the GC 'Test Motors' test after this patch. Before the patch, the TLE5206 board did not.

BarbourSmith:
It would be nice to understand what is going wrong and how we broke things, but I support this as a quick fix. I think we might encounter other weirdness down the road until we know what is happening so let's keep our eye on the underlying cause if we do

blurfl:
The TLE5206 chip is asserting its open-collector Error Flag, which we connect to ENx but don't pay attention to. (_pwmPin in Motor.cpp, note to self - needs to be initialized INPUT_PULLUP).
 When the movement commanded by Motor::write() is a start from a dead stop, the TLE5206 sees the motor windings as a dead short and asserts EF. The EF is cleared by toggling one or the other INx inputs. When the speed is less than full throttle, a pwm signal accomplishes this reset. Watching on a scope with pwm=254, after around 20 PWM pulses the motor has reached a speed where it no longer triggers EF.
 I haven't looked to see why it worked back before PR#489, but a guess is that back then Axis::detach() was called frequently (in detachIfIdle()?) which would have toggled the INx lines if they were HIGH.
  This patch would address the EF issue any time Motor::write() is called with max speed.
  I did find that another way to accomplish this is to put a 'soft start' in Axis::test() ahead of the second direction movement, like
```
    if (TLE5206) {
        // get the motor started moving
        motorGearboxEncoder.motor.directWrite(-20);
        maslowDelay(100);
    }
```
  I tested that approach with no load on the motors and it worked, I'd have to change the setup here to test with loads on the motors.
That would more tightly focus this patch on the 'Test Motors' issue. What do you think?

BarbourSmith:
Mmmm that makes more sense. So with a load on the motor it coasts less and so the issue is less likely to show up. If we put a delay in there does that solve the issue fully? I would feel more comfortable with a delay than a lockout of 255. Do you think this could be an issue while the machine is running, or does it seem like just a quirk with the test?

davidelang:
On Tue, 21 May 2019, BarbourSmith wrote:

> Mmmm that makes more sense. So with a load on the motor it coasts less and so 
> the issue is less likely to show up. If we put a delay in there does that 
> solve the issue fully? I would feel more comfortable with a delay than a 
> lockout of 255. Do you think this could be an issue while the machine is 
> running, or does it seem like just a quirk with the test?

If we are not going to monitor the error pin, then we need to avoid situations 
where this is critical. locking out 255 entirely is not that bad a thing.

David Lang

blurfl:
>If we are not going to monitor the error pin, then we need to avoid situations
where this is critical. locking out 255 entirely is not that bad a thing.

How about I put up a separate PR with the other approach and let the votes make the choice?

davidelang:
On Tue, 21 May 2019, Scott Smith wrote:

>> If we are not going to monitor the error pin, then we need to avoid situations
> where this is critical. locking out 255 entirely is not that bad a thing.
>
> How about I put up a separate PR with the other approach and let the votes make the choice?


If someone is willing to go to the effort of monitoring the pin and reacting 
properly (including that it happened), that is clearly the best thing to do. But 
if we don't have someone who is ready to do that, limiting it to 254 is a 
quick-and-dirty fix.

David Lang

blurfl:
>monitoring the pin and reacting

Unfortunately all the interrupts are committed for the encoders. In lieu of that where would be an appropriate place to trigger the check of the error flags? One possibility is at the end of execSystemRealtime() which runs from the main loop() as well as from maslowDelay();

davidelang:
On Tue, 21 May 2019, Scott Smith wrote:

>> monitoring the pin and reacting
>
> Unfortunately all the interrupts are committed for the encoders. In lieu of 
> that where would be an appropriate place to trigger the check of the error 
> flags? One possibility is at the end of execSystemRealtime() which runs from 
> the main loop() as well as from maslowDelay();

this stays high until cleared, so it doesn't need to be an interrupt.

how bad would it be to wait until the next position calculation loop cycle?

David Lang

blurfl:
>this stays high until cleared

 The Axis::test() function doesn't use position calculation, but it does call maslowDelay() which calls execSystemRealtime(). That might be a good enough place to check it.

davidelang:
this doesn't belong in the test subroutine, this is something that should be in 
place for any time the motor is being driven.

David Lang

blurfl:
>in place for any time the motor is being driven

Use your 👍 vote [here](https://github.com/MaslowCNC/Firmware/pull/501#issuecomment-494265798) for that 

or [here](https://github.com/MaslowCNC/Firmware/pull/503#issuecomment-494590135) for the patch that is confined to the 'Test Motors' issue.

PR #502, opened 2019-05-21
Added version 4

MarcinanBarbarzynca:
Here is link to thread
https://forums.maslowcnc.com/t/managing-motors-with-dual-l298n/9896
Here is link to new shield prepared under DIY and external H bridges for people that need more current for their motors.  
https://easyeda.com/polaski/maslow-cnc-shield-v2

I need to change attach version pins on board at the moment, so please write your suggestion on forum.

### DIFF
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -225,6 +225,39 @@ void   setupAxes(){
         aux8 = 46;
         aux9 = 47;
     }
+    else if(pcbVersion == 4){
+    //https://easyeda.com/polaski/maslow-cnc-shield-v2
+        //MP1 - Right Motor
+        encoder1A = 20; // INPUT
+        encoder1B = 21; // INPUT
+        in1 = 7;        // OUTPUT
+        in2 = 6;        // OUTPUT
+        enA = 5;        // errorFlag
+        //MP2 - Z-axis
+        encoder2A = 19; // INPUT
+        encoder2B = 18; // INPUT
+        in3 = 10;        // OUTPUT
+        in4 = 9;        // OUTPUT
+        enB = 8;        // errorFlag
+        //MP3 - Left Motor
+        encoder3A = 2;   // INPUT
+        encoder3B = 3;   // INPUT
+        in5 = 13;        // OUTPUT
+        in6 = 12;        // OUTPUT
+        enC = 12;        // errorFlag
+
+        //AUX pins
+        aux1 = 52;
+        aux2 = 50;
+        aux3 = 48;
+        aux4 = 44;
+        aux5 = 40;
+        aux6 = 36;
+        aux7 = 32;
+        aux8 = 30;
+        aux9 = 28;
+}
+
 
     if(sysSettings.chainOverSprocket == 1){
         leftAxis.setup (enC, in6, in5, encoder3B, encoder3A, 'L', LOOPINTERVAL);
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -1,18 +1,14 @@
 /*This file is part of the Maslow Control Software.
-
 The Maslow Control Software is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.
-
 Maslow Control Software is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
-
 You should have received a copy of the GNU General Public License
 along with the Maslow Control Software.  If not, see <http://www.gnu.org/licenses/>.
-
 Copyright 2014-2017 Bar Smith*/
 
 // This file contains system level functions and states
@@ -58,11 +54,8 @@ void  calibrateChainLengths(String gcodeLine){
 
 void   setupAxes(){
     /*
-
     Detect the version of the Arduino shield connected, and use the appropriate pins
-
     This function runs before the serial port is open so the version is not printed here
-
     */
 
 
@@ -203,7 +196,7 @@ void   setupAxes(){
         aux8 = -1;       // unconnected
         aux9 = -1;       // unconnected
     }
-    else if{ // (pcbVersion == 3) // TLE5206
+    else { // (pcbVersion == 3) // TLE5206
         //TLE5206 PCB v1.3 Detected
         //MP1 - Right Motor
         encoder1A = 20; // INPUT
@@ -237,8 +230,6 @@ void   setupAxes(){
         aux8 = 46;
         aux9 = 47;
     }
-    
-
 
     if(sysSettings.chainOverSprocket == 1){
         leftAxis.setup (enC, in6, in5, encoder3B, encoder3A, 'L', LOOPINTERVAL);
@@ -293,7 +284,6 @@ int getPCBVersion(){
             pinCheck &= B000111; // strip off the unstrapped bits
             TLE5206 = true;
             break;
-       
 }
     return pinCheck - 1;
 }
@@ -325,7 +315,7 @@ void setPWMPrescalers(int prescalerChoice) {
     TCCR4B &= ~prescalerEraser;   // this operation sets the three bits in TCCR4B to 0
     // now set those same three bits
 // ————————————————————————————–
-// TIMER 2       (Pin 9, 10)
+// TIMER 2       (Pin 9, 10)
 // Value  Divisor  Frequency
 // 0x01   1        31.374 KHz
 // 0x02   8        3.921 KHz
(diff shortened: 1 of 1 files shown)


BarbourSmith:
It looks like the lint check is complaining with error:

`System.cpp:240:5: error: 'else' without a previous 'if'`

Which I think is caused by the `else` statement on 206 being followed by the` if else` further down, I believe that the last statement needs to be `else`.

I really like what you have done with making it possible to use a stand alone L298N board. Are you want to use a custom board version instead of matching the inputs to one of the existing ones? If you feel like you really need to go that route it's OK by me.

blurfl:
I can't remember the specific reason, but I know that the present firmware is crafted to avoid using D11, D12 or D13 as a PWM pin. That could give you trouble. Look at Motor::write() to see what I mean. 

It would take some reworking the PCB, but using the same pin definitions for the L298 chip signals as the version 1 or version 2 board would make the software modifications _much_ easier. No changes would be needed for Motor.cpp, for instance, and that's where the tricky parts are.

davidelang:
The existing versions are based around if there is one direction pin and an 
enable pin to pulse or if there are different enable pins for each direction

David Lang

esspe2:
Maybe you have to fetch the last version of the firmware: with the PR #499, the test on the version 1.3 has been removed, so it has to be put back with a last `else` for your 4th version:

Before:

    else { // (pcbVersion == 3) // TLE5206
        //TLE5206 PCB v1.3 Detected
       ( ... )
    }

After:

    else if (pcbVersion == 3){ // TLE5206
        //TLE5206 PCB v1.3 Detected
        ( ... )
    }
    else { // (pcbVersion == 4){//version not assigned
        //DIY motor shield for external H bridges.  
        (...)
    }

MarcinanBarbarzynca:
> 
> 
> I can't remember the specific reason, but I know that the present firmware is crafted to avoid using D11, D12 or D13 as a PWM pin. That could give you trouble. Look at Motor::write() to see what I mean.
> 
> It would take some reworking the PCB, but using the same pin definitions for the L298 chip signals as the version 1 or version 2 board would make the software modifications _much_ easier. No changes would be needed for Motor.cpp, for instance, and that's where the tricky parts are.

Ok. Ill try to prepare new prototype for that :P

blurfl:
PCB creation, software crafting, learning to use github, there's a lot of tasks to making a new board, it's an exciting process and you're making a good contribution, but there's a bit more work to do before this is ready for a PR. ;)
 I think that the best way to do it is to wait until the whole thing - board and firmware - are made and tested and found to work, then create the PR. Before then, push your changes to your own fork, keep your own master current with the upstream master, and when all the coding is done with no more 'little touches' to take care of, then it will be time.

MarcinanBarbarzynca:
I need to know witch pins are connected to make it 1.2 Version board. Give me 2 H to make new PCB

esspe2:
Why is there already some mention of a 1.4 board before your mod?
You may have to rename yours to "1.5" because `getPCBVersion()`  comments are rather confusing.between 1.3 and 1.4, someone please correct me if I'm wrong.

I think that the CI (failed)check is based on the latest version, but you started from before the PR #499.
So it doesn't understand which version of line 206 you have, and doesn't put back the if.

Do something like : save all your work, git fetch origin, git branch board1.5, copy your work, make another PR. 
Or does some Git guru know a better way?

MarcinanBarbarzynca:
Here is DIY version 1.2 
https://easyeda.com/polaski/maslow-diy-1-2-shield-for-external-h-drivers

esspe2:
Here is some info for the pins of the documented board which is named "V1.2b":
From Config.h:
```
// Define version detect pins
#define VERS1 22
#define VERS2 23
#define VERS3 24
#define VERS4 25
#define VERS5 26
#define VERS6 27
```
Looking at:  https://raw.githubusercontent.com/MaslowCNC/Electronics/master/PowerDistributionBoardSchematic.PNG
- JP3 connects pins 22&23 to 5V , 
- JP4 connects pins 52&53 to GND.

By chance pins pullups are defined in `getPCBVersion()`, which means:
```
VERS1==1 (used)
VERS2==1 (used)
VERS3==1 (unconnected ??)
VERS4==1 (unconnected ??)
VERS5==1 (unconnected ??)
VERS6==1 (unconnected ??)
PIN52==0 (unused ??)
PIN53==0 (unused ??)
```
I replicated the JP3 JP4 connexions on my prototype, which seems to work, but I feel that something has to be done about those pins 52&53, maybe you should connect them to the ground too.

blurfl:
>Here is DIY version 1.2

That will make your life very much easier - you shouldn't need to make any changes to Motors.cpp with this new layout. 💯 This is a great example of a hardware solution ~solving~ eliminating a scaly software issue :P


>There is description witch pins connect to 5v and GND to get version 1.2

Here's my understanding:

```
On the original Maslow L298P boards, 
D22-D23 and D52-D53 on XIO were used to 
indicate the board revision number in binary. 
The software read these pins and detect the 
shield version.

#53-#52           #23-#22
-------           -------
GND-GND           GND-5V  -> rev.0001
GND-GND           5V-GND  -> rev.0002
GND-GND           5V-5V   -> rev.0003

Because D52 and D53 are used for 
SPI, newer boards should uses D22-D27.

#27-#26  #25-#24  #23-#22
-------  -------  -------
GND-GND  GND-5V   GND-GND -> rev 1.3. TLE5206
GND-GND  GND-5V   GND-5V  -> rev 1.4
GND-GND  GND-5V   5V -5V  -> rev 1.5  TLE9201 
```

You'll still need a version 1.4 definition, though, to get your AUX pin definitions. I didn't look to see whether your INCnA and INCnB pins match 1.2, but those will get handled by your 1.4 section as well. I don't think you would need any other changes to the firmware.

esspe2:
Thanks @blurfl , I was wondering what happened to the pins `#53-#52`.
Awww I see: duel between identification and SPI : "this town ain't big enough for the both of us".

Winner : SPI  !

But is it the end ? Maybe we still could use them in the version check AND afterwards for SPI if needed, I have some ideas.

MarcinanBarbarzynca:
Here is my version 1.4 board. 
https://docs.google.com/document/d/1dOIk3eNJ1dNO3xcWViaV02fj7BChyFAuJ0R7hZSsiss/edit?usp=sharing
https://easyeda.com/polaski/maslow-cnc-shield-v2
I updated pinouts 

Here is my version 1.2 board:
https://easyeda.com/polaski/maslow-diy-1-2-shield-for-external-h-drivers

blurfl:
Making great progress 👍🏻
If you're open to one more suggestion, instead of using the 5V supply pin of the Arduino, use the IOREF pin. The IOREF pin is internally tied to the 5V supply on Arduinos that use 5V, and tied to the 3.3V supply on Arduinos and compatibles that can be damaged by 5V on the inputs (Arduino Due, Teensy 3.6, Grand Central Metro). With that change, the boards would would work with the MaslowDue firmware with very few software changes. It's a hardware change that opens up software options.

MarcinanBarbarzynca:
Soo. What do You all think... Shall i get DIY version as 1.4? 
I connected version pins as blurfl suggested.
Now we have 1.4 with 9 AUX pins and 1.2 with 6 aux. 
I can't push my commit with 1.4 update :|.
On Arduino IDE i can upload main Maslow program, but when i try my own it gets some EEPROM error "internal compiler error: Segmentation fault".
When i replace version 3 with my code it works normally.

blurfl:
>Soo. What do You all think...

 Your v1.2 board looks like it will run using the stock releases of firmware, which is a big plus for supporting users who are more comfortable with woodworking than with firmware versioning.

I'd suggest that you develop your v1.4 board on your own fork of the firmware and wait to merge it here until you've done more with it there.

At this point, this PR only seems to modify some of the whitespace in the System.cpp file. Perhaps it should be closed?

MarcinanBarbarzynca:
Ye. I need to get rid my versioning. I will lock my self in 1.2 version for while, because I need to learn how to interact with github. 
I will prepare better guide, maybe video on "How to make Your Maslow 1.2 cnc shield for 15$" (10$ for h bridges + 5$ for other parts)

The version 1.4 witch 9 AUX is ready to get own place in System.cpp, but i wont post it (conflicts + warnings xD).  [https://docs.google.com/document/d/1dOIk3eNJ1dNO3xcWViaV02fj7BChyFAuJ0R7hZSsiss/edit](url) <The pinouts will stay the same for probably 6 months

blurfl:
There's a lot to learn with github :) Your v1.2 board looks like it will make it very easy for someone to make. Great contributions!

BarbourSmith:
Excellent progress! I think a write up on how to build a DIY shield cheaply would be an awesome thing for the community. I know you are not alone in wanting that so hopefully you can save someone else the hassle of having to work it all out.

Jumping way back in the conversation:

> I can't remember the specific reason, but I know that the present firmware is crafted to avoid using D11, D12 or D13 as a PWM pin. That could give you trouble. Look at Motor::write() to see what I mean.

Those pins use the same timers that we use to trigger some of our internal calculations to generate their PWM signals and so the PWM signals are not reliable on those pins. What you've done with using the 1.2 pinout is the right way to go. That is the best pinout we've found so far.

PR #503, opened 2019-05-21
Confine changes to Axis::test()

blurfl:
This patch only affects TLE5206 boards. This works around the symptom noted in issue #500, addressing it as a motor coasting issue. 

>The TLE5206 chip is asserting its open-collector Error Flag, which we connect to ENx but don't pay attention to. (_pwmPin in Motor.cpp, note to self - needs to be initialized INPUT_PULLUP).
When the movement commanded by Motor::write() is a start from a dead stop or while the motor is coasting and the direction abruptly changes, the TLE5206 can see the motor windings as a dead short and asserts EF.


davidelang:
we need to fix this everywhere, not just in the test() function.

BarbourSmith:
I like this dueling pull requests thing 😀 ... I think the other one is the proper solution because we don't want the machine to end up working differently during the test than during a cut

Issue #504, opened 2019-05-23
What board ID number to use?


blurfl:
I'm working on another board version, and I need to choose a version ID code that doesn't clash with any other project. I've provisionally used v1.5 thinking that by skipping one I would be safe. When @MarcinanBarbarzynca came up with a board that used ID v1.4, that didn't conflict. 👍
 I can stick with v1.5, leaving v1.4 for @MarcinanBarbarzynca's second board project. That will leave a gap in the sequence - how does the community feel about that?
 One reason to stake a claim on an ID number is that circuit boards get strapped for that number and it's a pain to cut/patch the PCBs after they are made.

BarbourSmith:
Way to be considerate of others!

I vote for taking v1.4, if you are ready to take it now. That way we won't have any skipped numbers which might confuse folks.

I think we should do a first come first served basis for pull requests merged into master so if you want to lock down a number just make a pull request to master with the pins you want to use.

davidelang:
On Wed, 22 May 2019, Scott Smith wrote:

> I'm working on another board version, and I need to choose a version ID code that doesn't clash with any other project.

what is it about your board that is going to require a unique ID?

If it can be made pin compatible with an existing board, that would be good.

We had a couple versions of the board from Bar because he found that different 
pins worked better for PWM, and then we had another version because the 
directiion controls worked differently (IIRC originally it was 
Enable_dir_1, Enable_dr_2 and the new chip with Dir, Enable), if your new board 
can fit either of these schemes, can we use the same ID? if not, what is 
different?

David Lang

blurfl:
Presently we have boards with the L298 motor driver that use three input pins to control it (one for master enable and two for individual output state - any of the three could act as the PWM signal) and boards with the TLE5206 that only has two control inputs, and chooses direction and speed by putting those in opposite states and putting a PWM signal on one of them.

>what is it about your board that is going to require a unique ID?

[...]

PR #505, opened 2019-05-23
Add comments to document the board versioning system

blurfl:


BarbourSmith:
Excellent documentation! I would thumbs up it three times if I could 👍 👍 👍

davidelang:
Could you add the explination about how the pins are used on the different 
versions (the same explination that you gave me about why you needed the new 
board id because the different chip types use different ways of signalling 
direction, speed, and enable)

David Lang

davidelang:
the code says which pins are input and output for each motor, but they don't 
describe the 'two pins to determine direction' vs 'one drection, one enable, 
either can be PWM' vs...

That's the type of info I think needs to appear in this file as well.

David Lang

blurfl:
>but they don't describe the 'two pins to determine direction' vs 'one drection, one enable, either can be PWM' vs...

That's better handled by reading the data sheets for the respective ICs. The complexities go beyond what's appropriate in a firmware comment. That's where it is appropriate to read the fine manual.

davidelang:
On Thu, 23 May 2019, Scott Smith wrote:

>> but they don't describe the 'two pins to determine direction' vs 'one drection, one enable, either can be PWM' vs...
>
> That's better handled by reading the data sheets for the respective ICs. The complexities go beyond what's appropriate in a firmware comment. That's where it is appropriate to read the fine manual.

in  that case, a link to the data sheets is appropriate (but I really thnk the 
simple explination you gave me would be good to have)

David Lang

blurfl:
I'm going to close this for now, as there is ongoing discussion in Issue #504. Best to wait for that discussion to finish, then implement those decisions.

BarbourSmith:
Ok, but if the discussion goes on too long and we need someone to just put their foot down, let me know. Ultimately I think having this new board exist is much more important than what version number it gets. I am really excited about this new chip. It is a very cool find.

blurfl:
> this new chip

[Here](https://www.digikey.com/product-detail/en/infineon-technologies/HBRIDGEKIT2GOTOBO1/HBRIDGEKIT2GOTOBO1-ND/5878318) is the $15 development board for the chip. Very slick! [Here](https://www.infineon.com/dgdl/Infineon-H-Bridge_Kit_2Go_UM-UM-v01_01-EN.pdf?fileId=5546d46250cc1fdf015133ee879d3696&ack=t) the manual for the dev kit (software is free, but haven't tried it).

PR #510, opened 2019-06-19
When (pinCheck >= 6) the board version matches the version strapping

blurfl:
Add comments to explain the zero-based version number reported from the 1-based gpio pin strapping as suggested in recent discussions:
[PR#502](https://github.com/MaslowCNC/Firmware/pull/502)
[Issue#504](https://github.com/MaslowCNC/Firmware/issues/504)
[PR#505](https://github.com/MaslowCNC/Firmware/pull/505)

 Beginning with board v1.6, the value from the gpio pins for new boards will be zero-based to match the number reported by the firmware and the silkscreen, as suggested here:
[Issue#504](https://github.com/MaslowCNC/Firmware/issues/504#issuecomment-495791288)
>Hidden gem: starting with v1.5, pin numbering matches version numbering, we are good at least until v1.15! After that some creativity will be necessary.
But for that, return pinCheck<5 ? pinCheck-1 : pinCheck;, in getPCBVersion()
to be discussed, of course!

[PR#502](https://github.com/MaslowCNC/Firmware/pull/502)
[PR#505](https://github.com/MaslowCNC/Firmware/pull/505)

There is a parallel/competing PR#511 which updates the comments without changing the version strapping - choose one or the other with your vote.

### DIFF
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -274,6 +274,32 @@ void   setupAxes(){
 }
 
 int getPCBVersion(){
+/*
+*       On the original Maslow L298P boards, 
+*     D22-D23 and D52-D53 on XIO were used to 
+*     indicate the board revision number in binary. 
+*     The software uses INPUT_PULLUP to read these pins 
+*     and detect the shield version. TLE5206 boards
+*     expanded the XIO pins to allow more board numbers.
+*       The value read from the gpio pins is/was 1-based,
+*     but the board version number silkscreened on those boards
+*     and reported by the firmware was zero-based - a source of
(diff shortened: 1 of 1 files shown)


blurfl:
PR#510 and PR#511 aren't compatible with each other, one or the other can be merged but not both. I couldn't see a way to be sure that only one would 'win' merging, so I guess I'll watch the votes and close the one with a lower total. Help with the vote count by giving one a 👍🏻and the other a 👎.

BarbourSmith:
Got it! I saw updating the comments first and thought updating comments is
always a good idea...thanks for pointing out that merging both isn't a good
idea. I will take my 👍 off of #511 when I next  get a stable internet
connection.

PR #511, opened 2019-06-19
Add comments about board versioning.

blurfl:
Explain the zero-based version number reported from the 1-based gpio pin strapping. Continue to use the same 1-based format for pin strapping.

Attempting to implement some of the suggestions in recent discussions:

[PR#502](https://github.com/MaslowCNC/Firmware/pull/502)
[Issue#504](https://github.com/MaslowCNC/Firmware/issues/504)
[PR#505](https://github.com/MaslowCNC/Firmware/pull/505)

There is a parallel/competing PR#51o which updates the comments and changes the version strapping for future boards - choose one or the other with your vote.

### DIFF
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -274,6 +274,27 @@ void   setupAxes(){
 }
 
 int getPCBVersion(){
+/*
+*       On the original Maslow L298P boards, 
+*     D22-D23 and D52-D53 on XIO were used to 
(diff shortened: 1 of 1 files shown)


blurfl:
PR#510 and PR#511 aren't compatible with each other, one or the other can be merged but not both. I couldn't see a way to be sure that only one would 'win' merging, so I guess I'll watch the votes and close the one with a lower total. Help with the vote count by giving one a 👍🏻and the other a 👎.

PR #512, opened 2019-06-19
Alarm if board version is unrecognized

blurfl:
• set all gpio pin numbers to 0
• Post an error dialog
• loop forever to cause a serial disconnect

Board-unrecognized Alarm message


BarbourSmith:
This is a great idea 👍👍

blurfl:
Credit for the idea goes to @esspe2, from his comment [here](https://github.com/MaslowCNC/Firmware/pull/499#issuecomment-493771373) 👍
>Either way, if I were to push my thoughts further, the runtime should stop and yell something like board unrecognized, or version pins configuration unknown.

PR #514, opened 2019-07-16
Limit PWM frequency value

blurfl:
The function that allows changing the PWM frequency needs some updating, to prevent invalid option values and to honor the PWM frequency limits of the driver chips.

- constrain choices for all driver chips to valid values 1..3
- update verboseDebug messages
- limit the PWM frequency to less than the driver chip upper limit:
-- the TLE5206 is limited to PWM frequencies below 1,000Hz

### DIFF
--- a/cnc_ctrl_v1/System.cpp
+++ b/cnc_ctrl_v1/System.cpp
@@ -358,23 +358,39 @@ int getPCBVersion(){
 
 //
 // PWM frequency change
-//  presently just sets the default value
-//  different values seem to need specific PWM tunings...
+//  constrain choices to valid values
+//  tailor the PWM frequency to less than the driver chip upper limit
 //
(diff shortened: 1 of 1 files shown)


cnc_ctrl_v1/System.cpp
```diff
@@ -358,23 +358,39 @@ int getPCBVersion(){
...
+                    Serial.println(F("4,100Hz"));
+                }
             break;
             case 3:
                 Serial.println(F("490Hz"));
             }
     #endif
+    // tailor the PWM frequency to the chip
+    if (TLE5206) {
+    // The upper limit to PWM frequency for TLE5206 is 1,000Hz
```
BarbourSmith: Thanks for always writing such easy to read code. Your comments are excellent.

PR #519, opened 2019-08-05
Adds directWrite(0) at end of B11

madgrizzle:
Adds a directWrite(0) at end so motors are stopped.  They would stop on their own after the next move command is issued (typically where slack is reintroducted) but for some reason, currently unknown, the chain length measurement B10 function isn't working consistently and the motors keep getting pulled.  B11 could use further tweeking (no need to send directWrites every iteration of the timing loop).

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -285,6 +285,12 @@ byte  executeBcodeLine(const String& gcodeLine){
             execSystemRealtime();
             if (sys.stop){return STATUS_OK;}
         }
+        if (gcodeLine.indexOf('L') != -1){
+            leftAxis.motorGearboxEncoder.motor.directWrite(0);
+        }
+        else{
+            rightAxis.motorGearboxEncoder.motor.directWrite(0);
+        }
         bit_false(sys.state,STATE_POS_ERR_IGNORE);
         return STATUS_OK;
     }


blurfl:
I've tested this and confirm that it corrects the issue.
To test, set a macro to 'B11', disconnect the chains, and run the macro. Without the patch the motors run indefinitely; with the patchthey stop after threed second as intended.
@BarbourSmith, I think we should commit this patch immediately, before the Robot time period - **Excellent work**  @madgrizzle 💯

davidelang:
definitely a longstanding bug. If it says it runs the motors for X time it 
should stop them after that time.

while you are thinking about this code, could you make it so that both motors 
can be given a speed in one statement?

David Lang

blurfl:
>while you are thinking about this code, could you make it so that both motors can be given a speed in one statement?
How about capturing that in a new issue? That would be a feature, this is a bug fix.

BarbourSmith:
Awesome work 👍👍👍👍

BarbourSmith:
Now how did we introduce this bug?

davidelang:
On Tue, 6 Aug 2019, BarbourSmith wrote:

> Now how did we introduce this bug?

I think it's always been there. But it was papered over by the fact that we only 
used this during calibration.

there was a change to logging in the holey triangular branch that caused some 
logs to be supressed, making it so GC never saw the completion of the command 
after this (the measurement) and so never issued the command following to 
release slack.

David Lang

PR #524, opened 2019-08-13
Initialize sys.feedrate to 400 mm/min

madgrizzle:
sys.feedrate was not being initialized and resulted in a feed rate of 1 mm/min when gcode was sent without a feed rate (if no other feed rate had been previously sent).  This sets the feedrate to 400 mm/min (half of the maximum feed rate).  It appears to me that the firmware works in metric so this should be appropriate .


blurfl:
This works as advertised, in inches as well as mm. 👍

A thought, though, how about using the sysSettings.maxfeed?

```    
sys.feedrate = sysSettings.maxFeed / 2.0;
```

davidelang:
I agree, half of the max feed setting would be better than have of the max that 
we think of as the machine limit.

madgrizzle:
modified.

PR #527, opened 2019-08-14
Add a basic template for pull requests

BarbourSmith:
This will create text which is auto populated in the pull request window when a new pull request is created. See the description of how this works here:

https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository

Let me know if we want to change the wording. This file can of course be itself edited by pull request so if we want to change the wording down the road it is always an option.

It will look like:

# What feature does this pull request add or what issue does this pull request fix?

# Does this firmware change affect kinematics or any part of the calibration process?
## d) Has this PR been tested on actual machine and/or in fake servo mode (indicate which or both)?

Thanks for contributing!

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -2,13 +2,16 @@ Thanks for contributing to The Maslow Firmware! You rock.
 
 Please let the community know some basic information about this pull request.
 
-# What feature does this pull request add or what issue does this pull request fix?
+## What does this pull request do?
+Does it add a new feature or fix a bug?
 
-# Does this firmware change affect kinematics or any part of the calibration process?
-## a) If so, does this change require recalibration?
-## b) If so, is there an option for user to opt-out of the change until ready for recalibration? If not, explain why this is not possible.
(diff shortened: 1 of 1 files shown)


blurfl:
How about asking the originator how to test to demonstrate the effect of the change? For example, from PR#519

```
To test, set a macro to 'B11', disconnect the chains, and run the macro.
 Without the patch the motors run indefinitely; with the patch they stop 
after three seconds as intended.
```

davidelang:
questions I would ask

what behavior is expected to change from this PR (i.e. fix a bug, no expected 
changes, new feature added, then provide details)

If this change affects calibration, can the user opt-out of recalibrating 
(recognizing that they don't get the benefit of the change, does an upgrade 
force them to recalibrate)

does this change require the firmware and GroundControl to both be changed (if 
so, link to the PR for the other component)

is there a way to test this change

BarbourSmith:
Great suggestions. I've updated the wording to read:

Does it add a new feature or fix a bug?

Thanks for contributing!

PR #529, opened 2019-08-20
Add Missing Kinematics.Init()

madgrizzle:
Any changes to values that are used during the kinematics::forward calculations should call kinematics.init() so sled position can be recalculated.  If its not, this can cause a "sled not keeping up" error, particularly after calibration.  If the sled is at an unadjusted position and a move is made, the move will be made with the adjusted parameters and that can exceed the positionErrorLimit.
Did I miss any?

Does it add a new feature or fix a bug?  **Fixes a bug**

**In a sense, it updates the sled position to utilize values just pushed to it**

**no**
**n/a**
**not required.. it eliminates the potential for a sled not keeping up error**
**fake servo only**

**Someone try it out by changing one of the values significantly and making the sled move.  Prior to this fix, if value change was significant enough, it would cause a sled not keeping up error.  This should fix it.**

Thanks for contributing!

### DIFF
--- a/cnc_ctrl_v1/Settings.cpp
+++ b/cnc_ctrl_v1/Settings.cpp
@@ -382,6 +382,7 @@ byte settingsStoreGlobalSetting(const byte& parameter,const float& value){
             break;
         case 37:
               sysSettings.chainSagCorrection = value;
+              kinematics.init();
               break;
         case 38:
               settingsSaveStepstoEEprom();
@@ -400,9 +401,11 @@ byte settingsStoreGlobalSetting(const byte& parameter,const float& value){
               break;
         case 40:
               sysSettings.leftChainTolerance = value;
+              kinematics.init();
               break;
         case 41:
               sysSettings.rightChainTolerance = value;
+              kinematics.init();
               break;
         case 42:
               sysSettings.positionErrorLimit = value;


blurfl:
Which values are you talking about? When would those be changed in normal use?

madgrizzle:
chainSagCorrection , leftChainTolerance, rightChainTolerance.  Normally these aren't changed so much that it results in a problem, but I think we saw it in v1.27 (@jess experienced it).  I've seen something like this happen in testing and never was able to figure it out, but by working on her issue I discovered what I think was the source of it.

davidelang:
On Tue, 20 Aug 2019, Scott Smith wrote:

> Which values are you talking about? When would those be changed in normal use?

calibration settings, they would not be changed in normal use, just when doing 
calibration, but they can make the calibration 'fail' and cause you to need to 
calibrate again or reset the chains

madgrizzle:
I believe the only impact is that the controller would kick a sled-not-keeping-up error.  The calibration should still be good.  Once you clear it, you are good to go.  This just eliminates that error from occurring.  Changes to other parameters (rotational radius, distance between motors, etc.) call kinematics::init, just not those three.  I think it was just an oversight.

blurfl:
> chainSagCorrection , leftChainTolerance, rightChainTolerance

So would a test sending '$37=nnnn.mm' '$40...' and '$41...' show the issue? Looking for a clear test here...

madgrizzle:
Yes.  The delta between the old and new value has to be significant enough to cause the new sled position to be more than your position error tolerance (2.0 mm default).

madgrizzle:
Once you get the initial sled-not-keeping-up alarm.  clear it and you shouldn't get it again.  When you do the fix, it shouldn't occur in the first place.

blurfl:
Sending the '$NN=' commands does trigger the Kinematics.init(), and no bad things seem to happen, but since those are calculated values GC recalculates them and sends back the original value the next time a '$$' report is sent.
 Haven't gone through a full recalibration to see whether the sled alarms are solved, though. 
 Moving the sled with changed '$NN=' values before GC restores them can trigger the alert with or without the patch, depending on the value. The alerts may be due to faulty measurements or flawed calculations?

madgrizzle:
"Sending the '$NN=' commands does trigger the Kinematics.init(), and no bad things seem to happen, but since those are calculated values GC recalculates them and sends back the original value the next time a '$$' report is sent."

Correct.. but there's no reason for a user to manually send the $NN= commands for that very reason.  

"Haven't gone through a full recalibration to see whether the sled alarms are solved, though.
Moving the sled with changed '$NN=' values before GC restores them can trigger the alert with or without the patch, depending on the value. The alerts may be due to faulty measurements or flawed calculations?"

It's not specifically a calibration issue.  The issue occurs during calibration because that's when chain sag and chain tolerance factors are recalculated.  If one manually changes their values in settings, it could occur then.

This is what I suspect is happening and I don't fully understand it, but I think I'm in the general ballpark... The firmware keeps track of the sled's position in sys.xPosition, sys.yPosition.  The kinematics::forward function is used to find the sled position from the chain lengths using the kinematics formula.  Let's say for a given set of chain lengths, the sled position is 0,0.  Then, you update the chainSagCorrection value.  With the current firmware, the sled is still at 0,0 because you haven't recalculated the sled position using kinematics::init (which calls kinematics::forward) but if you had called it, the sled position would have updated to 0,1 (just as an example).  Then you tell the sled to move 1 inch.  During coordinatedMove, calculations are based upon the current sys.xPosition, sys.yPosition (0,0) and those calculations result in a given _setPoint per step (or something like that).  Then kinematics::inverse is called and chain lengths get calculated.  But at that time, the chain lengths are based upon the new chainSagCorrection value.  What I believe is happening is... something like... the amount of chain length to be let out/reeled in for that step is calculated to be more than what the machine is capable of because its the steps were based upon an incorrect starting point (0,0 vs. 0,1).  I can't tell you precisely what is going on.. I don't understand it enough.. but I think that's where the demon lies.

blurfl:
Sorry, I wasn't clear. I don't question that it seems to be proper to run Kinematics.init() in these instances.
 I'm just commenting that it is being hard to create a test situation that repeatably fails without and succeeds with the patch. Having a test like that, which others can run and confirm, is reassuring.

madgrizzle:
"Sorry, I wasn't clear. I don't question that it seems to be proper to run Kinematics.init() in these instances."

I wasn't doubting that.. I think its intuitive that it should be run since its run with the "original" parameters are changed.  I agree it may be difficult to test because you have to create the conditions where it happens.. I tried to describe where I think the issue is occurring in case that helps.  I just know when @jess was trying to calibrate with v1.27 she was getting some really big numbers for chainTolerance (for unrelated reasons to this issue) and when they got pushed, after she tried to move it would raise that alarm (because of this issue).  I was able to duplicate that happening and when I added kinematics.init calls to those three items, it stopped happening.

PR #530, opened 2019-08-22
add B99 command to control FAKE_SERVO mode

blurfl:
FAKE_SERVO mode is useful for testing firmware and GC changes, or troublesome gcode files without needing to run the router around the board. It can be used on a Mega/motorControl pair with no motors attached.

  Use: 'B99 ON' turns FAKE_SERVO mode on, 'B99' with anything other than 'ON' (or nothing) turns it off.
  The current state of FAKE_SERVO is reported at startup and when it changes.
  The state is saved in EEPROM to persist over restarts, but is set off when the EEPROM is wiped, so off is the default state. 
  During FAKE_SERVO mode, the motors do not run. Instead computePID() spoofs the encoder value sent to the PID using a value that is +- 10% of the expected destination value. The cursor on the GC screen tracks the progress of the spoofed encoder values.

Files changed:
 cnc_ctrl_v1.ino:
 - add global flag to signal Axis::computePID() and Motor::write()
 - check for saved state at startup and report it

 Gcode.cpp:
 - add 'B99 ' command
 - - if the line contains 'ON' set FAKE_SERVO_STATE otherwise clear it
 - - report the state
 - - save the value of FAKE_SERVO_STATE in the high byte of EEPROM to persiste over restarts
 - - note that Setings::settingsWipe(SETTINGS_RESTORE_ALL) writes '0' to that location, which turns FAKE_SERVO off

 Axis.cpp:
 - at start of Axis::computePID(), check 'if (FAKE_SERVO_STATE != 0)' to determine whether to fake the encoder change. The check adds very few microseconds to this importrant function.

 Motor.cpp:
 - at the start of Motor::write() add a check for (FAKE_SERVO_STATE == 0) to determine whether to set the speed of the motor. This check adds only a few microseconds. Adding this check keeps the motors inoperative dureing FAKE_SERVO mode.

This adds a feature.

No, it does not change the kinematics.
It does not need recalibration.
The default state is off.
The change does not affect these, no changes are needed,
The PR has been tested on an actual machine _and_ in fake servo mode 😁.

Create two macros: 'B99' and B99 ON'
Start GC and connect, note the line 'FAKE_SERVO off' on the screen.
Test for correct motor operation.
Use the 'B99 ON' macro, note the line 'FAKE_SERVO on'
Test that the cursor moves on the screen as expected but the motors stay quiet.
Disconnect the serial connection and reconnect to demonstrate that the state of FAKE_SERVO persists over a restart.
Use the 'B99' macro, note the line 'FAKE_SERVO off'
Test for correct motor operation.

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -17,6 +17,7 @@
 
 
 #include "Maslow.h"
+// #include <EEPROM.h>
 
 void Axis::setup(const int& pwmPin, const int& directionPin1, const int& directionPin2, const int& encoderPin1, const int& encoderPin2, const char& axisName, const unsigned long& loopInterval)
 {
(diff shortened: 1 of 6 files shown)


BarbourSmith:
> The PR has been tested on an actual machine and in fake servo mode 😁.

😀 😀 😀 

Nice work :)

PR #539, opened 2020-02-25
Do not perform Z axis movement on G2 if Z is not specified in gcode line

gb0101010101:
- Do not perform Z axis movement on G2 if Z is not specified in gcode line.
- Set Z1 and Z2 to NAN if not provided. 
- Update arc() to deal with NAN Z. 
- Make Z calculations happen the same way as X&Y before movementUpdate().
- Add Z end move same as X & Y.

## Does it add a new feature or fix a bug?
Removes unnecessary z axis calculation and faulty movement. This fixes a problem in with Z axis not reaching target depth when G2 is executed as zxais.read was getting current z position whilst it was still in motion and using that for the G2 Z. This code lets the previous Z movement complete. 

No.
#### d) Has this PR been tested on actual machine and/or in fake servo mode (indicate which or both)?
Fake servo only. Will hopefully test on actual machine this weekend.

### G2 code without Z
```
G0 G40 G90  
G21  
G0 X0 Y0 Z0  
G1 Z-8  
G2 X37.93360 Y5.515 I43.2022 J129.306
G0 Z0  
```

### G2 code with Z
```
G0 G40 G90 G17  
G21  
G0 X0 Y0 Z0  
G1 Z-8  
G2 X37.93360 Y5.515 I43.2022 J129.306 Z-12 
G0 Z0  
```

### DIFF
--- a/cnc_ctrl_v1/GCode.cpp
+++ b/cnc_ctrl_v1/GCode.cpp
@@ -20,6 +20,7 @@ Copyright 2014-2017 Bar Smith*/
 
 #include "Maslow.h"
 #include <EEPROM.h>
+#include "math.h"
 
 maslowRingBuffer incSerialBuffer;
 String readyCommandString = "";  //KRK why is this a global?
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Awesome work @gb0101010101! 👍 👍 

I don't have hardware to test this on, but it will get my vote if someone can confirm that it works as expected in hardware.

I like how you cleaned up the comments along the way 👍

BarbourSmith:
This is a great change, I vote that we re-open this as soon as someone tests it in hardware. I just don't want to risk a change to the master branch of the firmware without having proof in hardware.

davidelang:
this could just be that nobody had a chance to test it and provide the up-vote 
needed here.

David Lang

BarbourSmith:
Yeah, I think that is exactly what happened 😢

PR #540, opened 2020-03-03
gb0101010101-g2-arc-z-optional: Do not perform Z axis movement on G2 …

gb0101010101:
Testing create pull request from same branch for @bar.

Same pull request as before but now tested on physical machine.


BarbourSmith:
It looks like for some reason the robot can't see a second pull request from the same branch 😢 

I'm really going to have to overhaul how that system works. Thanks for highlighting some of the issues with how it works now.

PR #541, opened 2020-05-15
Add support for G2 & G3 commands that use radius R instead of I & J.

gb0101010101:
## What does this pull request do?
Adds support for G2 & G3 commands that use radius R instead of I & J. 
e.g. `G02 X2.0 Y6.0 R2.0`

The code only executes if R is provided in G2 or G3 command. If R, I & J are provided then I & J will be ignored.

Previously this was not supported and code execution would fail if the command was used.

Calculations are copied from MarlinFirmware
https://github.com/MarlinFirmware/Marlin/pull/4840/commits/c2744d8a8bdd35a824a1017434a7b3007debdc1e 

## Does this firmware change affect kinematics or any part of the calibration process?
No.

## How can this pull request be tested?
Turn on verbose debugging to see calculated points in console output.
This has not been tested on physical machine. I will not have access to a machine for a while.

Rendered output:
http://www.helmancnc.com/cnc-g02-circular-interpolation-clockwise-cnc-milling-sample-program/
### Test G2 code
```
G90 G00 X-2.0 Y-1.0 
G01 X0 Y0 F30.0 ; point A 
Y4.0 ; point B 
G02 X2.0 Y6.0 R2.0 ; point C 
G01 X8.0 ; point D 
G02 X9.0 Y2.268 R2.0 ; point E 
G01 X0 Y0 ; point A 
G00 X-2.0 Y-1.0
```
### Test G3 code
```
G90 G00 X-2.0 Y-1.0 
G01 X0 Y0 F30.0 ; point A 
Y4.0 ; point B 
G03 X-2.0 Y6.0 R2.0 ; point C 
G01 X-8.0 ; point D 
G03 X-9.0 Y2.268 R2.0 ; point E 
G01 X0 Y0 ; point A 
G00 X-2.0 Y-1.0
```

## GroundControl/WebControl
This PR will need to match changes to GroundControl and WebControl so that they render the commands correctly.
WebControl PR https://github.com/madgrizzle/WebControl/pull/137

PR #544, opened 2020-05-19
Add support for G2 & G3 commands that use radius R instead of I & J.

gb0101010101:
Resubmitting PR #541


gb0101010101:
Tested on physical machine by @Orob-Maslow with pictures in https://github.com/WebControlCNC/WebControl/pull/137

PR #549, opened 2021-06-06
TB6643 v1.5 support

EBS-Maslow:
1- Added support for EBS shield TB6643 v1.5
2- Added acceleration control for "test motors/encoders" to avoid triggering the TLE5206/TB6643 overcurrent alarm/cut-off while testing (https://forums.maslowcnc.com/t/motor-wont-stop-turning/16680/8)
3- Fix for a small bug in Motor.cpp line 119 which prevented the brake for being applied.
4- Added laserOff to sys.stop and systemReset for safety
5- maxFeed bumped up to 900 (~35ipm)

### DIFF
--- a/cnc_ctrl_v1/Axis.cpp
+++ b/cnc_ctrl_v1/Axis.cpp
@@ -302,19 +302,24 @@ void   Axis::test(){
     //print something to prevent the connection from timing out
     Serial.print(F("<Idle,MPos:0,0,0,WPos:0.000,0.000,0.000>"));
     
-    int i = 0;
+ 
     double encoderPos = motorGearboxEncoder.encoder.read(); //record the position now
-    
-    //move the motor
-    motorGearboxEncoder.motor.directWrite(255);
-    while (i < 1000){
-        i++;
-        maslowDelay(1);
-        if (sys.stop){return;}
+
+
+    //ramp up to full speed
+    for (int ramp = 1 ; ramp <= 255; ramp += 15) {
+    motorGearboxEncoder.motor.directWrite(ramp);
+    maslowDelay(30);
+    if (sys.stop){return;}
     }
-    //stop the motor
-    motorGearboxEncoder.motor.directWrite(0);
-    
+    maslowDelay(3000);
+    //ramp down to stop
+    for (int ramp = 241 ; ramp >= 1; ramp += -15) {
+    motorGearboxEncoder.motor.directWrite(ramp);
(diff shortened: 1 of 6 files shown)


cnc_ctrl_v1/Axis.cpp
```diff
@@ -302,19 +302,24 @@ void   Axis::test(){
...
-    //move the motor
-    motorGearboxEncoder.motor.directWrite(255);
-    while (i < 1000){
-        i++;
-        maslowDelay(1);
-        if (sys.stop){return;}
+
+
+    //ramp up to full speed
+    for (int ramp = 1 ; ramp <= 255; ramp += 15) {
```
BarbourSmith, line 310: Nicely done

cnc_ctrl_v1/System.cpp
```diff
@@ -414,18 +453,28 @@ int getPCBVersion(){
...
             TLE9201 = false;
             break;
-        case B000110:
+        case B000110: //  v 5
+            pinCheck &= B000110; // 110 for 6 
+            TB6643 = true;
+            TLE5206 = false;
+            TLE9201 = false;
+            break;
+        case B000111:  //v 1.6 AND V 1.7 (1.7 is a place holder and will not exist because 1.6 uses the same pins)
```
BarbourSmith, line 470: Why can't 1.7 exist?

BarbourSmith:
I'm seeing this as closed, is that right? These changes seem great. I dilly dallied on merging them because I don't really have the hardware to test them and I was trying to think up a way to, but I think they all look good and the only thing for it is to merge them and handle any issues which come up (which I wouldn't expect there to be any) when they come up.

cnc_ctrl_v1/System.cpp
EBS-Maslow: I'm following orob lead here, i know he is testing a new TLE9201 shield and contemplating on encoding it as v1.6.
Going from base 1 to 0 encoding there is no way to not lose a #version, if we make the change from pinCheck <7 instead 8 we could have v1.7 shield but not v1.6

EBS-Maslow:
Yes, I'm sorry, wanted to do some changes and due to my lack of github experience just closed it. Will do a new pull request.

cnc_ctrl_v1/System.cpp
BarbourSmith: Gotcha that makes sense

PR #550, opened 2021-06-09
Adding TB6643 v1.5 support

EBS-Maslow:
1- Add support for EBS TB6643 v1.5 shield
2- Added pwm ramp to "test motors/encoders" function to avoid triggering the overcurrent alarm on the TB/TLE shields while testing.
3- Bumped up maxFeed to 900 (~35imp)
4- Fix a small bug in Motor.cpp line 119 that prevented from applying the motor brake when speed = 0
5 - Added lasserOff to system.stop & systemReset for safety


BarbourSmith:
I don't have much ability to test this other than to verify that it compiles...but it compiles so that's the best I can do 😁

PR #554, opened 2022-07-25
Controller Shield v1.8b Hardware Update

redchairlabs:
Previous hardware version 1.8a had pin asignments that resulted in the PWM Prescalers being incorectly assigned. Hardware v1.8b will fix this issue but also requires a coresponding firmware update. This update does not affect kinematics or calibration.


BarbourSmith:
Very nice!! Thanks for keeping it up to date!

